PR #34, opened 2018-02-06
Django project skeleton

ilyavw:
This is the initial django skeleton. 

There are two major components: config and account. Config is your standard configuration module that contains the settings. Account will be the django app that deals with account related tasks. You'll also notice the stubbed out django-admin command.

### DIFF
--- a/.pyup.yml
+++ b/.pyup.yml
@@ -6,4 +6,4 @@ requirements:
   - tox.ini:
       pin: False
   - requirements/test.txt:
-      ping: False
+      pin: False
--- a/.pyup.yml
+++ b/.pyup.yml
@@ -5,3 +5,5 @@ search: True
 requirements:
   - tox.ini:
       pin: False
+  - requirements/test.txt:
+      ping: False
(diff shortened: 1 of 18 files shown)


.pyup.yml
```diff
@@ -5,3 +5,5 @@ search: True
 requirements:
   - tox.ini:
       pin: False
+  - requirements/test.txt:
+      ping: False
```
infinitewarp: should be `pin`?

cloudigrade/account/management/commands/add_account.py
```diff
@@ -0,0 +1,8 @@
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    help = 'This adds aws accounts'
+
+    def handle(self, *args, **options):
+        self.stdout.write(self.style.SUCCESS('AWS Account Added... probably'))
```
infinitewarp: Since this command doesn't actually do anything yet, let's change the message to a more _obvious_ dummy placeholder. Maybe something like "AWD Account command placeholder message". I'm also a big fan of `# FIXME` comments to make it super obvious to readers that things aren't working as expected.

PR #35, opened 2018-02-07
Replace coveralls with codecov.io

ilyavw:
It seems like coveralls is broken, if it does succeed in generating coverage it takes 1+ hours to do it.


ilyavw:
@elyezer Both really, it is very customizable so you can present a lot more data.

PR #36, opened 2018-02-12
Save customer ID

### LINKED ISSUE #18: Save customer account ID
As a Doppler admin, I want to access customers’ AWS accounts via AWS APIs.

This means we give a customer *our* AWS Account ID, the customer creates a role with our policy JSON, then… TBD? See also: [Tutorial: Delegate Access Across AWS Accounts Using IAM Roles ](https://docs.aws.amazon.com/IAM/latest/UserGuide/tutorial_cross-account-with-roles.html)

## Acceptance Criteria

- [x] Verify I can give CLI Doppler a customer's AWS ARN.
- [x] Verify that the DB table contains:
  - [x] TBD… but probably customer account ID
  - [x] ARN

## Tech assumptions

- Since we expect the future service to be a web app, we will go ahead and start this using Django even though effectively this is a "simple" CLI.
- Store things in a local sqlite DB using Django migrations to build the schema.
- “Verification” here can be done by having a dev reveal the contents of the file/DB since good APIs won’t arrive until later.
- We do not care at this time about deduping, collision checking, updating, etc.
- The account ID from AWS is *effectively* our primary key or at least a sort-of unique index.
- All columns in our tables that are references to AWS-based IDs should be strings.
- For now, we plan to access the AWS APIs by using our Doppler account, and the credentials for that account are expected to live in the environment at `~/.aws/`. This will later be mounted/substituted/etc. per running environment.
- All tables in our DB should have auto-populated columns for time created and modified (names TBD, @infinitewarp can provide example of this for Django model)

ilyavw:
This will take a customers ARN as CLI argument, extra account ID, and store both pieces of information in the database. Also included docstrings and improved linting and coverage reporting.

Closes #18


cloudigrade/account/models.py
```diff
@@ -0,0 +1,15 @@
+"""Cloudigrade Account Models."""
+from django.db import models
+
+from util.models import BaseModel
+
+
+class Account(BaseModel):
+    """Account model."""
+
+    account_id = models.DecimalField(
```
infinitewarp: `BigIntegerField` instead? Then we don't need the first two kwargs.

cloudigrade/account/models.py
```diff
@@ -0,0 +1,15 @@
...
+        max_digits=12,
+        decimal_places=0,
+        db_index=True
+    )
+    account_arn = models.CharField(max_length=256)
```
infinitewarp: We probably want an index on this field too since we'll have lookups on it.

cloudigrade/account/migrations/0001_initial.py
```diff
@@ -0,0 +1,27 @@
...
+    initial = True
+
+    dependencies = [
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Account',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
```
adberglund: For a migration you need to specify the primary key autofield, but not for the model itself? Basically, I need to learn about migrations.

cloudigrade/account/tests/test_management_commands.py
```diff
@@ -0,0 +1,24 @@
+"""Collection of tests for Account management commands."""
+from io import StringIO
+from django.core.management import call_command
+from django.utils.translation import gettext as _
+from django.test import TestCase
+
+
+from ..models import Account
```
infinitewarp: Absolute imports > relative imports

cloudigrade/account/models.py
adberglund: When you have 2 columns, this seems trivial, but I've seen a directory of models grow to 100+ models. It can be really nice to have a short comment on each column giving some specifics. Say if there's a check constraint, it's a good place to put the possible/expected values and what they mean. 

Here I would consider just a quick note that these are currently AWS specific.

cloudigrade/util/models.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class BaseModel(models.Model):
+    """Abstract model to add automatic created_at and updated_at fields."""
+
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        abstract = True
```
infinitewarp: Below here, we may want to add:

    ordering = ('created_at',)

so we have a default ordering for future models, but I don't feel too strongly about this right now.

cloudigrade/account/models.py
adberglund: Will this always be numeric, or do we want a string here to accommodate other clouds?

cloudigrade/account/models.py
```diff
@@ -0,0 +1,15 @@
...
+from util.models import BaseModel
+
+
+class Account(BaseModel):
+    """Account model."""
+
+    account_id = models.DecimalField(
+        max_digits=12,
+        decimal_places=0,
+        db_index=True
```
adberglund: This table may be small enough that this index won't be used.

cloudigrade/account/models.py
infinitewarp: On the other hand, Django automatically creates an `id` field for all models and this explicit field is redundant. Let's just drop it unless you have another reason I don't see.

cloudigrade/account/models.py
infinitewarp: and `unique=True`… though I'm not sure if adding that will automagically give it an index.

cloudigrade/account/models.py
infinitewarp: Maybe we should even call the column `aws_account_id`?

cloudigrade/account/models.py
infinitewarp: but Soon™?

cloudigrade/account/models.py
infinitewarp: @ilyavw and I discussed offline… It makes sense to keep this as a 12-digit decimal because that's how AWS defines its account ID: a 12-digit integer.

infinitewarp: @ilyavw and I discussed offline… let's just leave this as-is because we're going to have to restructure the whole model anyway when we extend support beyond AWS, and it makes sense to make AWS-specific distinctions at that time.

.codecov.yml
```diff
@@ -0,0 +1,19 @@
+coverage:
```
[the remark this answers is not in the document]
ilyavw: Generally, yes. #18 was kind of a problem since it encompassed both the skeleton project creation and implementation of features. The story really should have been split, but since it wasn't, you'll see both things here.

cloudigrade/account/management/commands/add_account.py
```diff
@@ -1,9 +1,21 @@
+"""Management command for storing AWS ARN credentials in the database."""
 from django.core.management.base import BaseCommand
+from django.utils.translation import gettext as _
```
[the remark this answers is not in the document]
ilyavw: What would you like to know?

cloudigrade/account/migrations/0001_initial.py
ilyavw: The migration is a generated file, if you do not specify a primary key in your model, django will provide one for you.

cloudigrade/account/models.py
ilyavw: I think AWS is kinda assumed at this point. account_arn wouldn't make much sense for gcp or azure, when we add support for other clouds I see this table having some drastic changes, but I don't think it makes sense currently to add any aws specific distinctions yet.

cloudigrade/account/models.py
adberglund: Postgres will definitely give you a b-tree index on unique constraint.

cloudigrade/account/management/commands/add_account.py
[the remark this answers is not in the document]
adberglund: Nope, I didn't know django housed gettext as a util. I will now use it. Just a comment.

cloudigrade/account/models.py
adberglund: Yeah, maybe. Even with a few thousand entries I wouldn't be shocked to see the query planner ignore the index as we bang our heads trying to get it to use the index.

cloudigrade/account/models.py
adberglund: Sure, I stand by my usefulness of column comments though. As you add more people looking at the code base it's just a service to newcomers.

PR #37, opened 2018-02-12
Display list of currently running EC2 instances for a given ARN

### LINKED ISSUE #23: See inventory of currently running instances
As a Doppler admin, I want to see an inventory of any instances that were already running when a customer began using Doppler (at account setup) so that I can track hourly usage for instances that are long-running (i.e. not typically rebooted).

## Acceptance Criteria

- [ ] Verify that I see a list of instances and AMIs that are currently powered on for a given customer.
- [ ] Verify that for each instance we see, specifically we see the:
  - [ ] `InstanceId` (e.g. `i-0228fecebf85774d5`)
  - [ ] `InstanceType` (e.g. `t2.micro`)
  - [ ] `ImageId` (e.g. `ami-0b1e356e`)
  - [ ] `NetworkInterfaces.SubnetId` (e.g. `subnet-d87eddb0`)

## Tech assumptions

- We will simply iterate through all currently known AWS regions and run EC2 "describe instances".

infinitewarp:
Closes #23.


infinitewarp:
WIP. Not yet ready for review/merge.

infinitewarp:
Still pending TODO: I want to carve out tests just for the new `utils.aws` functions instead of relying implicitly on the tests for the `check_instances` command.

Otherwise, things are shaping up.

cloudigrade/instance/management/commands/check_instances.py
```diff
@@ -0,0 +1,28 @@
...
+
+from django.core.management.base import BaseCommand
+from django.utils.translation import ugettext as _
+
+from util import aws
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
```
ilyavw: Should this actually be a separate management command? This check is something that runs at account setup, so would it make sense for the `account.management.commands.add_account` to run a function that would do this check instead?

cloudigrade/instance/management/commands/check_instances.py
```diff
@@ -0,0 +1,28 @@
+"""Management command for checking currently-running AWS EC2 instances."""
+import logging
+
+from django.core.management.base import BaseCommand
+from django.utils.translation import ugettext as _
```
ilyavw: `ugettext` doesn't really do anything in python3, we should keep it consistent as `gettext` as everywhere else.

cloudigrade/instance/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
+
+
+class InstanceConfig(AppConfig):
+    name = 'instance'
```
ilyavw: I'm not 100% on if we need an app for instance functionality, as I don't expect us to have too much instance specific functions. We should talk more about this.

cloudigrade/instance/management/commands/check_instances.py
```diff
@@ -0,0 +1,28 @@
...
+
+class Command(BaseCommand):
+    """Command to display a list of running instances for a given ARN."""
+
+    help = _('Display a list of running instances for a given ARN.')
+
+    def add_arguments(self, parser):
+        """Add an --arn argument for use by this command."""
+        parser.add_argument(
+            '--arn', dest='arn', required=True,
```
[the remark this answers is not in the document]
infinitewarp: Good point. I was trying to think if there would be a case where this would expand to have more arguments in the future, but I don't have any now and it's easy enough to change later if needed. I'll simplify this down.

cloudigrade/instance/management/commands/check_instances.py
infinitewarp: Funny timing; I was talking this morning with @adberglund about that same idea but for his work. In this case, I was thinking of our fine-grained approach we had discussed for user stories with step 1 being "do thing independently" and step 2 being "merge disparate things into single command", but after checking I see that this story _doesn't_ have that delineation in its requirements. So, yeah, I think it makes sense to roll it up into the same command.

cloudigrade/instance/management/commands/check_instances.py
infinitewarp: Good catch. Old habits die hard. ☠️

cloudigrade/instance/apps.py
ilyavw: Quick summary of our conversation: Roll instance functionality into account app. General plan is to have three main apps, account, scheduler, and analyzer (well, util too). Account handles basic account setup and prep, scheduler deals with queues and scans, and analyzer actually processes the log files.

cloudigrade/account/tests/test_management_commands.py
```diff
@@ -1,11 +1,15 @@
...
 from io import StringIO
+from unittest.mock import patch
+
 from django.core.management import call_command
-from django.utils.translation import gettext as _
 from django.test import TestCase
+from django.utils.translation import gettext as _
 
-
+from account.management.commands.add_account import aws
```
ilyavw: are we importing from the right place here?

infinitewarp: Yes, with the way the mocking framework and Python imports work, you generally want to patch relative to where the code that needs the thing you're patching, not from where the thing originally lived. This is actually a common question (and can be a source of pain when getting started with patching), and they have a note about it here:

https://docs.python.org/3/library/unittest.mock.html#where-to-patch

> The basic principle is that you patch where an object is _looked up_, which is not necessarily the same place as where it is defined. A couple of examples will help to clarify this.

Issue #22, opened 2018-02-05
Ensure customer's IAM role/policy has required privileges enabled


infinitewarp:
As a Doppler admin, I want to programmatically access customers accounts so I can retrieve EC2 and CloudTrails data.

## Acceptance Criteria

- [ ] Verify that each required privilege is enabled by testing related API calls upon customer setup
- [ ] Verify that when a required privilege is not enabled, the CLI user sees an appropriate error message and causes account setup/save to abort

## Tech assumptions

- For now, we will provide to the customer a JSON blob (defining an IAM policy) and instructions to manually add an IAM user and role so Doppler can access that customer's AWS account.

kdelee:
Ok, I am a bit confused on what this story is about.

Is the intent to add a feature where CLI users can save customer access credentials (IAM) to Doppler, and upon creation Doppler immediately validates these credentials and ensures that they have the required level of access?

If so, I think the STORY should be "As a Doppler admin, I can save IAM to doppler to access EC2 and CloutTrails data at a later time"
Then "The IAM are automatically validated as having proper access privileges at creation time" is an acceptance criteria.

PR #42, opened 2018-02-15
21 confirm role access

adberglund:
I expect verify_account_access to change immediately following this with #22


cloudigrade/util/aws.py
```diff
@@ -49,7 +50,11 @@ def get_credentials_for_arn(arn):
 
     """
     sts = boto3.client('sts')
-    session_name = 'temp-session'  # TODO Does this need to be a funciton arg?
```
adberglund: I didn't meant to wipe your TODO, and will add it back in if wanted. My impression is that the session name matters very little. I could be very, very wrong.

cloudigrade/account/management/commands/add_account.py
```diff
@@ -29,9 +29,13 @@ def handle(self, *args, **options):
...
         account_id = aws.extract_account_id_from_arn(arn)
 
-        # TODO AWS account permissions verification goes here.
-        found_instances = aws.get_running_instances(arn)
-        Account(account_arn=options['arn'], account_id=account_id).save()
-
-        self.stdout.write(self.style.SUCCESS(_('ARN Info Stored')))
-        self.stdout.write(_(f'Instances found include: {found_instances}'))
+        is_verified = aws.verify_account_access(arn)
+        if is_verified:
```
ilyavw: why not just `if aws.verify_account_access(arn)`, it doesn't look like we use `is_verified` anywhere else.

cloudigrade/util/aws.py
infinitewarp: I'm okay with this! 👍

infinitewarp:
Since you have a merge commit in the history here, please be sure to use "**Squash and merge**" when closing this.

ilyavw:
Or to add to what Brad said above, squash it out manually if you'd like to preserve some other commits.

PR #44, opened 2018-02-16
Save inventory of currently running instances

infinitewarp:
Example command usage and output:

```
$ rm db.sqlite3 && \
  ./manage.py migrate 2>&1 >/dev/null && \
  ./manage.py add_account $ARN
ARN Info Stored
Running instances: {'us-east-2': [<Instance i-02ef56ef0b3029aba>]}
```

Then check the database:

```sql
select * from account_account;
select * from account_instance;
select * from account_instanceevent;
```


infinitewarp:
And example database contents after running the management command:

```
$ sqlite3 db.sqlite3 "select * from account_account"
1|2018-02-16 20:42:51.731406|2018-02-16 20:42:51.731471|518028203513|arn:aws:iam::518028203513:role/grant_cloudi_to_372779871274

$ sqlite3 db.sqlite3 "select * from account_instance"
1|2018-02-16 20:42:51.735397|2018-02-16 20:42:51.735606|i-02ef56ef0b3029aba|us-east-2|1

$ sqlite3 db.sqlite3 "select * from account_instanceevent"
1|2018-02-16 20:42:51.736174|2018-02-16 20:42:51.736199|power_on|2018-02-16 20:42:51.735997|subnet-930eeae9|ami-0b1e356e|1
```

cloudigrade/util/aws.py
```diff
@@ -10,6 +10,37 @@
...
+    See also:
+        https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_InstanceState.html
+
+    """
+
+    pending = 0
+    running = 16
+    shutting_down = 32
+    stopping = 64
+    stopped = 80
```
ilyavw: We should just include all the values here instead of leaving one out.

```
    0 : pending
    16 : running
    32 : shutting-down
    48 : terminated
    64 : stopping
    80 : stopped
```

cloudigrade/account/models.py
```diff
@@ -13,3 +14,54 @@ class Account(BaseModel):
...
+    )
+    region = models.CharField(
+        max_length=256,
+        null=False,
+        blank=False,
+    )
+
+    def __repr__(self):
+        """Get repr of this Instance."""
+        return f"<Instance {self.ec2_instance_id}>"
```
ilyavw: ' > "

cloudigrade/util/aws.py
infinitewarp: Good catch. I'm not sure how I missed that.

cloudigrade/account/models.py
infinitewarp: I think it's time to install `flake8-quotes` and run it with the build. I'll do that now!

cloudigrade/util/aws.py
```diff
@@ -10,6 +10,38 @@
...
+
+        Args:
+            code (int): The code from an EC2 Instance state.
+
+        Returns:
+            bool: True if we consider the instance to be running else False.
+
+        """
+        return code in (cls.running.value,
+                        cls.shutting_down.value)
```
ilyavw: Should we consider shutting down as a running instance? There will be no event written to CloudTrail when it does shut down, as such I think we may have to treat a shutting down state as already turned off.

cloudigrade/account/management/commands/add_account.py
```diff
@@ -24,17 +28,52 @@ def add_arguments(self, parser):
         """Require an ARN to be provided."""
         parser.add_argument('arn', help=_('Granted Role ARN'))
 
+    @transaction.atomic
```
adberglund: Do we want atomic to be our default behavior?

infinitewarp: Generally, I would say yes, _but_ I'm wary of leaving open long-running transactions due to the unknown duration of external API calls. That's why I shuffled the code to do all the API calls first and _then_ do the various saves.

If we want to push the lever to do atomic everywhere, I would be fine with that as long as we're careful like this about how we do API calls separated from write operations.

cloudigrade/util/aws.py
infinitewarp: Hmm… I haven't actually tried this myself, but I had (maybe wrongly) expected CT to send an event when the shutdown was complete. Where did you find your info?

ilyavw: Testing. CloudTrails doesn't log states of machines. It logs actions, so there is a log made when you as a user make a call to `stopInstance()`, which is when it'll go into the `shutting_down` state.

infinitewarp: Good to know! I'll pull this out.

cloudigrade/account/management/commands/add_account.py
adberglund: I can easily go either way. Let's do what you have and if it looks like every time we're using the atomic decorator we can assess down the line.

infinitewarp:
@ilyavw and @adberglund I've added saving of ec2 instance type. I'm just taking it as the bare string from the API response. Do you think this is OK, or should this be an enum on our end?

adberglund:
@infinitewarp I think it's fine as a string. I don't want to maintain that enum, and they can add types at any time. There's already a sea of them https://aws.amazon.com/ec2/instance-types/

Issue #27, opened 2018-02-05
Read CloudTrails metadata for on/off events


infinitewarp:
As a Doppler admin, I want to know when an EC2 instance is powered on and off so that I can bill customers appropriately.

## Acceptance Criteria 

- [ ] Verify that a CLI user sees the event metadata that CloudTrails delivers regarding instances powering on and off.
- [ ] Verify that a CLI user does *not* see events that had an error state; these should have been discarded.

## Tech assumptions

- CLI arguments include:
  - SQS queue URL
- Event metadata should include:
  - Event type (start/power on, stop/power off)
  - Event time
  - Instance ID
  - Instance size (e.g. micro)
  - AMI ID
  - subnet
- We need to be careful and explicit about checking for errors
  - For example, we may have a "stop instance" event but the action didn't really complete for some reason.
  - We probably want so just check for a value in the `Records[x].errorCode` of the event response.
- We still have very crude or nonexistent error handling.

kdelee:
Have you considered allowing users to create named collections of accounts/s3 buckets? 

If these are long hash type IDs I think this could be pretty annoying to have to specify at the command line over and over. I'd rather specify them once and then be able to refer to them by name.

adberglund:
I have done the bulk of the util work for interacting with SQS and S3 so that this log reader/analyzer can just process logs. I will post a PR for the util work that enables this shortly.

PR #49, opened 2018-02-20
Pre 27 sns s3 utils

adberglund:
This util work can facilitate the story requirements for #27

### DIFF
--- a/cloudigrade/util/aws.py
+++ b/cloudigrade/util/aws.py
@@ -150,10 +150,10 @@ def verify_account_access(arn):
     Check role for proper access to AWS APIs.
 
     Args:
-            arn (str): Amazon Resource Name to use for assuming a role.
+        arn (str): Amazon Resource Name to use for assuming a role.
 
-        Returns:
-            bool: Whether role is verifed.
+    Returns:
+        bool: Whether role is verifed.
 
     """
     role_session = get_assumed_session(arn)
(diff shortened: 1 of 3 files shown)


cloudigrade/util/aws.py
```diff
@@ -220,3 +222,88 @@ def _verify_policy_action(session, action):
...
+
+    object_bytes = s3_object.get()['Body'].read()
+
+    if compression == 'gzip':
+        content = gzip.decompress(object_bytes).decode('utf-8')
+    elif compression is None:
+        try:
+            content = object_bytes.decode('utf-8')
+        except UnicodeDecodeError as e:
+            logger.error(_(e))
```
ilyavw: Not sure if we can honestly expect gettext to know what to do with an exception message. ¯\_(ツ)_/¯

infinitewarp: @ilyavw I think your shrug guy dropped this: \\

adberglund: I'm not even being consistent, either as right above this I am decoding without try/except. In keeping with our "very minimal to no error handling" I could just expect it to be UTF-8 at this point? Alternatively I can replace with `logger.error(_('Content is not UTF-8 encoded'))` and duplicate the try/except above this.

ilyavw: Quite a bit of work goes into fixing that and the lean `¯\\\_(ツ)\_/¯`

ilyavw: Maybe throw in a small test for that branch while you're at it? ;)

PR #53, opened 2018-02-20
Check to make sure that all required policy actions are allowed.

ilyavw:
* Update session handling
* Update how other methods get a session
* Pull in test dependencies in local requirements file

### DIFF
--- a/cloudigrade/util/aws.py
+++ b/cloudigrade/util/aws.py
@@ -1,6 +1,7 @@
 """Helper utility module to wrap up common AWS operations."""
 import decimal
 import enum
+import json
 import logging
 
 import boto3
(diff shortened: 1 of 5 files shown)


cloudigrade/util/aws.py
```diff
@@ -122,49 +145,76 @@ def get_running_instances(arn):  # TODO filter
...
-        boto3.Session: A temporary session tied to a customer account
+        bool: Whether role is verified.
 
     """
-    if credentials is None:
-        credentials = get_credentials_for_arn(arn)
+    success = True
+    for action in cloudigrade_policy['Statement'][0]['Action']:
+        if not _verify_policy_action(session, action):
+            success = False
```
infinitewarp: You probably want `return False` here to break the iteration. Otherwise, if you have one action fail and the _next_ one succeeds, the whole function returns `True`.

PR #69, opened 2018-02-23
add readme for developers

infinitewarp:


Makefile
```diff
@@ -15,3 +15,6 @@ clean:
 reinitdb:
 	rm -rf $(TOPDIR)/db.sqlite3
 	$(PYTHON) $(PYDIR)/manage.py migrate --settings=config.settings.local
+
+unittest:
+	$(PYTHON) $(PYDIR)/manage.py test --settings=config.settings.local account util
```
ilyavw: In my personal experience this will fail unless the `--settings` is specified last. You sure you don't have a env var set with the settings file making this work?

Issue #28, opened 2018-02-05
Save CloudTrails metadata for on/off events


infinitewarp:
As a Doppler admin, I want customer EC2 instance power on/off events stored in my database so that I can easily report on them later.

## Acceptance Criteria 

- [ ] Verify that on/off records are stored in the DB
- [ ] Verify that all of the attributes from #27 are included in the DB models
  - [ ] Event type (start/power on, stop/power off)
  - [ ] Event time
  - [ ] Instance ID
  - [ ] Instance size (e.g. micro)
  - [ ] AMI ID
  - [ ] subnet

## Technical assumptions

- The DB models should already exist per a story completed in the previous sprint. See @infinitewarp if you have questions.

kdelee:
Can you enumerate what all data will be saved in conjunction with this on/off data -- i.e. what is the primary key? How are we going to merge this with other data?

adberglund:
A caveat I've discovered while working on #27

`RunInstances`: Create new instance(s), this has more info
`StartInstances`: Start a stopped pre-existing instance, this just has instance-id
`StopInstances`: Stop a pre-existing instance, this just has instance-id
`TerminateInstances`: Bye forever, instance. Not sure yet what data this has, let's find out.

When we go to save the data, we can either 
a) see if we've seen the instance before in our DB and fill in the info with what we have there, and if not call EC2 describe_instances() or 
b) always call EC2 describe_instances()

PR #70, opened 2018-02-23
27 read cloudtrail events

adberglund:


cloudigrade/analyzer/tests/test_management_commands.py
```diff
@@ -0,0 +1,184 @@
...
+                }
+            ]
+        }
+        mock_message = helper.generate_mock_sqs_message(
+            mock_queue_url,
+            json.dumps(mock_sqs_message_body),
+            mock_receipt_handle
+        )
+
+        mock_cloudtrail_log = mock_cloudtrail_log = {
```
adberglund: I can break this one into 3 separate tests to appease _the_ codecov.

cloudigrade/analyzer/admin.py
```diff
@@ -0,0 +1,4 @@
+"""Module to expose our models in Django admin."""
```
infinitewarp: Delete this file until we actually need it.

cloudigrade/analyzer/models.py
```diff
@@ -0,0 +1,4 @@
+"""Cloudigrade Analyzer Models."""
```
infinitewarp: Delete this file until we actually need it.

cloudigrade/analyzer/management/commands/analyze_log.py
```diff
@@ -0,0 +1,158 @@
+"""Management command for analyzing AWS CloutTrail Logs."""
+import json
+
+from django.core.management.base import BaseCommand
+from django.utils.translation import gettext as _
+
+# from account.models import Account, Instance, InstanceEvent
```
infinitewarp: Remove comment-out code.

cloudigrade/analyzer/tests/test_management_commands.py
```diff
@@ -0,0 +1,184 @@
...
+        mock_ec2_ami_id = str(uuid.uuid4())
+        mock_instance_type = 't2.nano'
+
+        expected_instances = [(mock_account_id, mock_instance_id, mock_region)]
+        expected_instance_events = [
+            (mock_instance_id, 'power_on', mock_occured_at, mock_subnet,
+             mock_ec2_ami_id, mock_instance_type)
+        ]
+
+        expected_out = _('Found instances: {i} and events {e}'.format(
```
infinitewarp: I know this is still a problem all over our code, but if/when we _actually_ start gettexting our strings, we want to only gettext the unformatted parts. So, this would need to become:

```
expected_out = _('Found instances: {i} and events {e}').format(i=expected_instances, e=expected_instance_events)
```

cloudigrade/analyzer/management/commands/analyze_log.py
```diff
@@ -0,0 +1,158 @@
...
+            bucket = parsed_message[0]
+            key = parsed_message[1]
+            logs.append(aws.get_object_content_from_s3(bucket, key))
+
+        # Parse logs for on/off events
+        for log in logs:
+            if log is None or log == '':
+                continue
+            instances, instance_events = self._parse_log_for_ec2_events(log)
+            result = 'Found instances: {i} and events {e}'.format(
```
infinitewarp: wrap `'Found instances: {i} and events {e}'` in `_(`...`)`

cloudigrade/analyzer/management/commands/analyze_log.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # Grab the object contents from S3
+        for parsed_message in parsed_messages:
+            bucket = parsed_message[0]
+            key = parsed_message[1]
+            logs.append(aws.get_object_content_from_s3(bucket, key))
+
+        # Parse logs for on/off events
+        for log in logs:
+            if log is None or log == '':
```
infinitewarp: Maybe simplify to `if not log`?

On second thought, maybe flip the condition around to `if log` and delete the `continue`.

cloudigrade/analyzer/admin.py
ilyavw: Lets remove the file until we actually have models to put in here.

cloudigrade/analyzer/management/commands/analyze_log.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # TODO: Save the messages to the DB
+        # TODO: If we haven't seen this instance before we'll
+        #       need to run describe_instances on it.
+        #       Start and stop requests don't have all the
+        #       info we need.
+
+        # aws.delete_message_from_queue(queue_url, messages)
+
+    def _parse_sqs_message(self, message):
```
infinitewarp: Part of me feels like this belongs in the `aws` module as a more general-purpose function next to `receive_message_from_queue`, but I'm OK with leaving it here until something else needs to read from the queue.

cloudigrade/util/tests/helper.py
```diff
@@ -92,3 +93,23 @@ def generate_dummy_describe_instance(instance_id=None, image_id=None,
         'SubnetId': subnet_id,
     }
     return mock_instance
+
+
+def generate_mock_sqs_message(id, body, receipt_handle):
```
infinitewarp: Avoid redefining the Python builtin `id`. Let's call this `message_id`.

cloudigrade/analyzer/management/commands/analyze_log.py
```diff
@@ -0,0 +1,158 @@
...
+
+            for item in ec2_info:
+                instance_id = item.get('instanceId')
+                instance_type = item.get('instanceType')
+                ec2_ami_id = item.get('imageId')
+                subnet = item.get('subnetId')
+
+                instances.append((account_id, instance_id, region))
+
+                # Subnet, ec2_ami_id, and instance_type will be None
```
ilyavw: Why? its kinda important for us to know the those things when the instance starts/stops. We need to get that info somehow I think.

adberglund: As to why, I can't speak for AWS; they just don't include the info for those calls. But, I agree. There's a TODO above to run an describe_instances call to get that info. I was in a bit of a Catch 22 and decided this is a task for the next story where we actually save info the DB. 

I can re-post the comment here, but I commented directly on #28 with the two paths I think we can take there.

cloudigrade/analyzer/management/commands/analyze_log.py
adberglund: I always get myself into weird negative check land when if log is obviously simpler.

cloudigrade/analyzer/management/commands/analyze_log.py
adberglund: I can be swayed. I put it here because there can be different message types in SQS. This function isn't, in its current form, generalized enough to accomodate that, but it *IS* a util for what we want now, so I'll gladly put it there now. If we ever see another kind of SQS message, we can fix it then.

cloudigrade/util/tests/helper.py
adberglund: 🤢 what have I done!?!?

Issue #33, opened 2018-02-05
JSON report of hourly RHEL usage for customer during time period


infinitewarp:
As a Doppler admin, I want a report of RHEL usage per customer so that I can ensure that we are billing them appropriately.

## Acceptance Criteria 

- [ ] Verify that I get JSON in the CLI output that contains total hourly usage aggregated by the combination of RHEL version and EC2 instance size for a customer

## Tech assumptions

- Probably query using Django ORM instead of raw SQL
- CLI input parameters:
  - start date/time
  - end date/time
  - AWS Account ID from setup/on-boarding step
- The particulars of the JSON structure are not yet strictly defined

infinitewarp:
Demo: https://asciinema.org/a/UAXzRbpboW5dnOa2qMW1zYLKr

PR #73, opened 2018-02-26
More helpers and helper tests

infinitewarp:
This is in support of #33 and #72, but since the actual report testing is taking a while, I wanted to go ahead and get some eyes on these new helpers.

### DIFF
--- a/cloudigrade/account/tests/helper.py
+++ b/cloudigrade/account/tests/helper.py
@@ -1,0 +1,105 @@
+"""Helper functions for generating test data."""
+import random
+import uuid
+
+from account.models import Account, Instance, InstanceEvent
+from util import aws
+from util.tests import helper
+
+
+def generate_account(arn=None, account_id=None):
+    """
+    Generate an Account for testing for random ARN and AWS Account ID.
+
+    Args:
+        arn (str): Optional ARN for the Account.
+        account_id (decimal.Decimal): Optional AWS Account ID.
+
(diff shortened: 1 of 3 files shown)


cloudigrade/account/tests/helper.py
```diff
@@ -0,0 +1,105 @@
...
+        region = random.choice(helper.SOME_AWS_REGIONS)
+
+    return Instance(
+        account=account,
+        ec2_instance_id=ec2_instance_id,
+        region=region
+    )
+
+
+def generate_instance_events(instance, powered_times, ec2_ami_id=None,
```
adberglund: `powered_times` here referred to as `powerd_on_times` in docstring.

cloudigrade/account/tests/test_helper.py
```diff
@@ -0,0 +1,128 @@
...
+
+from account.models import Account, Instance, InstanceEvent
+from account.tests import helper
+from util.tests import helper as util_helper
+
+
+class AccountHelperTest(TestCase):
+    """Test helper functions test case."""
+
+    def test_generate_generate_account_default(self):
```
adberglund: is the double `generate` here and in fhte next test intentional?

cloudigrade/account/tests/helper.py
infinitewarp: good 👀.

PR #76, opened 2018-02-27
Move -dev readme info into the main readme, update with required brew…

ilyavw:
… install


docker-compose.yml
```diff
@@ -3,6 +3,10 @@ version: '3'
 services:
   db:
     image: postgres:alpine
+    ports:
+      - "5432:5432"
+    volumes:
+      - /tmp/postgresql:/var/lib/postgresql
```
infinitewarp: For the local side, `${TMPDIR-/tmp}` is safer instead of the absolute `/tmp`.

infinitewarp: Actually… I'm not sure how that translates into the yaml world, but the suggestion is still sound. ¯\\\_(ツ)\_/¯

ilyavw: Docker won't use it by default, it requires extra configuration. 

https://docs.docker.com/docker-for-mac/osxfs/#namespaces

infinitewarp:
A nice demo of this would include something like:

```
make clean
make reinitdb

ARN_WITH_RUNNING="arn:aws:iam::518028203513:role/grant_cloudi_to_372779871274"
./cloudigrade/manage.py add_account ${ARN_WITH_RUNNING}
```

PR #83, opened 2018-03-01
28 save on off events

adberglund:


cloudigrade/analyzer/management/commands/analyze_log.py
```diff
@@ -136,3 +170,36 @@ def _is_on_off_event(self, record, valid_events):
...
+            #       DB multiple times. Consider caching.
+            account = Account.objects.get(account_id=account_id)
+            instance, __ = Instance.objects.get_or_create(
+                account=account,
+                ec2_instance_id=instance_id,
+                region=instance_dict['region'],
+            )
+            for event in instance_dict['events']:
+                occurred_at_dt = datetime.datetime.strptime(
+                    event['occurred_at'], '%Y-%m-%dT%H:%M:%SZ'
```
infinitewarp: Just me not knowing AWS's internals here…

Does AWS always use the `Z` time zone in its timestamps?

cloudigrade/analyzer/management/commands/analyze_log.py
```diff
@@ -136,3 +170,36 @@ def _is_on_off_event(self, record, valid_events):
...
+                ).replace(tzinfo=tz.tzutc())
+                instance_event = InstanceEvent(
+                    instance=instance,
+                    event_type=event['event_type'],
+                    occurred_at=occurred_at_dt,
+                    subnet=event['subnet'],
+                    ec2_ami_id=event['ec2_ami_id'],
+                    instance_type=event['instance_type'],
+                )
+                instance_event.save()
```
infinitewarp: You can drop the `instance_event.save()` if you change the previous line from `instance_event = InstanceEvent(` to `InstanceEvent.objects.create(`.

cloudigrade/analyzer/tests/test_management_commands.py
```diff
@@ -5,39 +5,46 @@
...
         out = StringIO()
         mock_queue_url = 'https://sqs.queue.url'
         mock_receipt_handle = str(uuid.uuid4())
-        mock_account_id = str(helper.generate_dummy_aws_account_id())
         mock_instance_id = str(uuid.uuid4())
         mock_region = 'us-east-1'
         mock_event_type = 'RunInstances'
-        mock_occured_at = datetime.datetime.utcnow().isoformat()
+        now = datetime.datetime.utcnow()
+        # Work around for sqlite handling of microseconds
```
infinitewarp: Oooh… I don't know if I've experienced or noticed this implied sqlite microsecond problem. What specifically did this work around?

cloudigrade/analyzer/tests/test_management_commands.py
```diff
@@ -79,27 +88,158 @@ def test_command_output_success(self):
...
+            self.assertEqual(event.ec2_ami_id, mock_ec2_ami_id)
+            self.assertEqual(event.instance_type, mock_instance_type)
+
+    def test_command_output_success_lookup_ec2_attributes(self):
+        """Test processing a CloudTrail log with missing instance data."""
+        out = StringIO()
+        mock_queue_url = 'https://sqs.queue.url'
+        mock_receipt_handle = str(uuid.uuid4())
+        mock_instance_id = str(uuid.uuid4())
+        mock_region = 'us-east-1'
```
infinitewarp: maybe `random.choice(util_helper.SOME_AWS_REGIONS)`

cloudigrade/analyzer/management/commands/analyze_log.py
```diff
@@ -93,26 +98,55 @@ def _parse_log_for_ec2_events(self, log):
...
+        Returns:
+            dict: EC2 instance data for the event.
+
+        """
+        ec2_ami_id = log_record.get('imageId')
+        instance_id = log_record.get('instanceId')
+        instance_type = log_record.get('instanceType')
+        subnet = log_record.get('subnetId')
+
+        if subnet is None \
```
infinitewarp: If any of them are `None`, we squash _all_ of them?

Is there ever a case where we'd get and want to keep some from the `log_record`? Or is it safe to assume that if any are missing, they're probably all missing and we need fetch from the instance?

cloudigrade/analyzer/management/commands/analyze_log.py
```diff
@@ -136,3 +170,36 @@ def _is_on_off_event(self, record, valid_events):
...
+
+        Args:
+            instances (dict): Instance data, including a list of events.
+
+        """
+        for instance_id, instance_dict in instances.items():
+            account_id = instance_dict['account_id']
+            # TODO: Don't pull the same account record from the
+            #       DB multiple times. Consider caching.
+            account = Account.objects.get(account_id=account_id)
```
ilyavw: I realize its a todo, but is there a reason you didn't pull lines 184 and 187 outside the for loop? Doesn't look like they have to be in the loop.

cloudigrade/analyzer/management/commands/analyze_log.py
adberglund: Realistically, I think if any of them is `None` all of them are. Those values _can't_ be none on an active EC2 instance, so we'll need to get the data regardless.

cloudigrade/analyzer/management/commands/analyze_log.py
adberglund: Fairly certain, yes. Is there a safer way to approach this?

cloudigrade/analyzer/tests/test_management_commands.py
adberglund: Specifically that `utcnow()` returns a datetime with microseconds, and when you read back the entry from the DB it cuts off at seconds. I assumed/thought it was sqlite, but maybe Django? My datetimes were definitely not equal going in and coming out though.

cloudigrade/analyzer/management/commands/analyze_log.py
adberglund: Only because I'm a baddy. 🙁

cloudigrade/analyzer/management/commands/analyze_log.py
infinitewarp: An alternative would be to use `dateutil` because it tries to be _clever_ about parsing date strings, but it we're fairly certain and don't expect ambiguity, then let's stick with what you have! 👍 

For completion sake, here's an example of what `dateutil`'s behavior looks like:
```
>>> from dateutil import parser
>>> parser.parse('2000-01-01T01:23:45')
datetime.datetime(2000, 1, 1, 1, 23, 45)
>>> parser.parse('2000-01-01T01:23:45Z')
datetime.datetime(2000, 1, 1, 1, 23, 45, tzinfo=tzutc())
>>> parser.parse('2000-01-01T01:23:45GMT')
datetime.datetime(2000, 1, 1, 1, 23, 45, tzinfo=tzutc())
>>> parser.parse('2000-01-01T01:23:45UTC')
datetime.datetime(2000, 1, 1, 1, 23, 45, tzinfo=tzutc())
>>> parser.parse('2000-01-01T01:23:45+0500')
datetime.datetime(2000, 1, 1, 1, 23, 45, tzinfo=tzoffset(None, 18000))
>>> parser.parse('2000-01-01T01:23:45-5')
datetime.datetime(2000, 1, 1, 1, 23, 45, tzinfo=tzoffset(None, -18000))
```

adberglund: I haven't used the parser before, so thanks for shedding 💡 on it. Putting in my toolbox. I do ❤️ `tz` and `relativedelta`.

PR #97, opened 2018-03-05
add favicon.ico

infinitewarp:
because @ilyavw asked for it

it currently just serves out of http://localhost:8000/static/favicon.ico


infinitewarp:
We got it working on @ilyavw's machine by purging his Docker with great 🔥, but alas…

PR #99, opened 2018-03-05
Create customer account via RESTful HTTP request

infinitewarp:


cloudigrade/account/models.py
```diff
@@ -85,3 +88,44 @@ def product_identifier(self):
 
         """
         return f'RHELX-{self.ec2_ami_id}-{self.instance_type}'
+
+
+def create_initial_instance_events(account, instances_data):
```
ilyavw: I don't think this belongs here, as this has nothing to do with defining a model.

cloudigrade/account/tests/test_serializers.py
```diff
@@ -0,0 +1,101 @@
...
+                patch.object(aws, 'get_running_instances') as mock_get_running:
+            mock_assume_role = mock_boto3.client.return_value.assume_role
+            mock_assume_role.return_value = mock_role
+            mock_verify.return_value = True
+            mock_get_running.return_value = mock_instances
+            serializer = AccountSerializer()
+
+            result = serializer.create(mock_validated_data)
+            self.assertIsInstance(result, Account)
+            # call_command('add_account', mock_arn, stdout=out)
```
ilyavw: Please remove commented out code.

cloudigrade/account/models.py
infinitewarp: I agree, and I was pondering a name for a module where this would belong when I decided to just stick it here for the time being. Any new module name suggestions, @ilyavw ?

ilyavw: `account.util`?

infinitewarp: 👌

cloudigrade/account/util.py
```diff
@@ -0,0 +1,47 @@
+"""Various utility functions for the account app."""
+import collections
```
ilyavw: I'm seeing some inconsistency with this. Sometimes there is a blank space after the doc string, sometimes there isn't. We should settle on one (as I don't think its defined in the PEP) and have it be uniform through our repo.

cloudigrade/account/models.py
```diff
@@ -1,4 +1,5 @@
 """Cloudigrade Account Models."""
+
```
ilyavw: Another empty space being added, refer to the other comment.

cloudigrade/account/management/commands/add_account.py
```diff
@@ -1,12 +1,11 @@
 """Management command for storing AWS ARN credentials in the database."""
-import collections
```
ilyavw: Another empty space, refer to the other comment.

cloudigrade/account/util.py
infinitewarp: At a quick glance, I didn't see an option to make flake8 opinionated about spacing after the module doc, and pyflakes (which is normally _much_ more opinionated about _many_ more things) also didn't report on it either way. So, for now I just manually removed the spaces since that appears to be our prevailing convention.

PR #103, opened 2018-03-06
80 auth

adberglund:
*Shockingly* little actual change to make this happen.


Makefile
```diff
@@ -3,6 +3,10 @@ PYTHON	= $(shell which python)
 TOPDIR  = $(shell pwd)
 PYDIR	= cloudigrade
 
+USERNAME= $(shell $(PYTHON) $(PYDIR)/manage.py shell -c \
+	"""from django.contrib.auth.models import User; \
+		print(User.objects.filter(is_superuser=True).first().username);""")
```
ilyavw: Yea I don't really know how I feel about this. If we are creating the user manually, I think we can manage to provide the username and password when logging in too.

adberglund: Sure thing. I was on the fence, between ease of `just give me a working key for now` and doing it for the exact user. I actually did it both ways, so I can push this up in a minute.

infinitewarp: Also, it doesn't work as implemented because I think it needs `--settings=config.settings.local`. Without that, it fails like this:
```
Traceback (most recent call last):
  File "/Users/brasmith/.virtualenvs/cloudigrade/lib/python3.6/site-packages/environ/environ.py", line 271, in get_value
    value = self.ENVIRON[var]
  File "/Users/brasmith/.virtualenvs/cloudigrade/bin/../lib/python3.6/os.py", line 669, in __getitem__
    raise KeyError(key) from None
KeyError: 'DJANGO_SECRET_KEY'
```

Issue #81, opened 2018-02-28
Get usage report via RESTful HTTP request


infinitewarp:
As a Doppler user, I want to run the "report_usage" command via HTTP commands so that I can see the usage data in a web browser.

## Acceptance Criteria 

- [ ] Verify that I can `GET /api/v1/report/?account_id={ACCOUNT_ID}&start={START}&end={END}` with a customer's Amazon account ID for `{ACCOUNT_ID}`, and the report start and end dates as ISO-8601 datetimes for `{START}` and `{END}`, and I receive the same report output as JSON that I would have received via the CLI.
- [ ] Verify that requesting for an ARN that doesn't exist returns a 404.

## Tech assumptions

- We don't need to worry about things like no-cache headers for now. We'll just trust our clients to make new requests as needed since we don't know exactly in what conditions reports will be idempotent.
- The start/end dates inclusivity/exclusivity behavior matches the CLI.

infinitewarp:
Demo: https://asciinema.org/a/168464

PR #111, opened 2018-03-08
expose report_usage command behavior as restful api

infinitewarp:
implements #81

### DIFF
--- a/cloudigrade/account/reports.py
+++ b/cloudigrade/account/reports.py
@@ -6,7 +6,7 @@ import operator
 
 from django.db import models
 
-from account.models import InstanceEvent, Instance
+from account.models import Account, Instance, InstanceEvent
 
 
 def _calculate_instance_usage(instance_events, start, end):
@@ -65,6 +65,10 @@ def get_hourly_usage(start, end, account_id):
         magical combination of Red Hat product version and AWS instance type.
 
     """
+    # Verify that we can get the requested account before proceeding.
+    if not Account.objects.filter(account_id=account_id).exists():
+        raise Account.DoesNotExist()
+
     # Get the nearest event *before* the reporting period for each instance.
     instances_before = Instance.objects.filter(
         account__account_id=account_id,
@@ -107,4 +111,4 @@ def get_hourly_usage(start, end, account_id):
         product_identifier, time_running = calculated_usage
         product_times[product_identifier] += time_running
 
-    return product_times
+    return dict(product_times)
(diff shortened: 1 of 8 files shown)


cloudigrade/account/reports.py
```diff
@@ -65,6 +65,9 @@ def get_hourly_usage(start, end, account_id):
         magical combination of Red Hat product version and AWS instance type.
 
     """
+    # Verify that we can get the requested account before proceeding.
+    Account.objects.get(account_id=account_id)
```
ilyavw: We should probably use `Account.objects.filter(account_id=account_id).exists()`, it should be a little faster since we are just checking for the presence of the object. Additionally, maybe we could add `if not` and throw an exception if the account does not exist?

cloudigrade/account/tests/test_serializers.py
```diff
@@ -1,14 +1,14 @@
...
 
 from django.test import TestCase
 from django.utils.translation import gettext as _
 from rest_framework import serializers
 
-from account.management.commands.add_account import aws
 from account.models import Account, Instance, InstanceEvent
-from account.serializers import AccountSerializer
+from account.serializers import AccountSerializer, ReportSerializer, reports
+from util import aws
```
ilyavw: If we are testing serializers, shouldn't this come from `account.serializers`?

cloudigrade/config/urls.py
```diff
@@ -4,10 +4,11 @@
 from django.urls import path, include
 from rest_framework import routers
 
-from account.views import AccountViewSet
+from account.views import AccountViewSet, ReportViewSet
 
 router = routers.DefaultRouter()
 router.register(r'account', AccountViewSet)
+router.register(r'report', ReportViewSet, base_name='report')
```
ilyavw: Is there a reason for setting the `base_name` here? It doesn't look like we defined a custom `get_queryset` method in the `ReportViewSet `.

infinitewarp: The `base_name` is necessary because `ReportViewSet` doesn't have a custom `get_queryset`, and I didn't give it a custom `get_queryset` because I couldn't think of a meaningful queryset to put there since all the fun model work for reporting happens a few steps away in `get_hourly_usage`.

ilyavw:
We decided to not remove `extract_sqs_message` from `util.aws` since we expect it to be used again soon™. Opened #114 to keep an eye on it.

Issue #66, opened 2018-02-22
More aggressive code quality checking


infinitewarp:
As a Doppler developer, I want even more aggressive code quality checking so that I can rest assured that I am writing clean(er?) code.

## Acceptance Criteria 

- [ ] Verify that additional quality checks are being run during CI.
- [ ] Verify that failures in the new checks also fail the pipeline.

## Tech assumptions

- Add and configure one or more of:
  - [Pylint](https://github.com/PyCQA/pylint)
  - [Prospector](https://github.com/landscapeio/prospector)
  - [Bandit](https://github.com/openstack/bandit)
- Side note: Prospector is made by the Landscape.io people, and [Landscape.io](https://landscape.io/) appears to be a dead service. Its website is usually down, but at least Prospector is open source and independently runnable.
- Included in this story is the effort required to evaluate the default rules for the chosen tools and adjust those rules to fit how we want to develop.

ilyavw:
After much discussion, we've come to the conclusion that pylint (the best of the three noted here) will not work for us.

PR #117, opened 2018-03-12
More aggressive quality checking

ilyavw:
#66

In addition to adding pylint this also introduces some changes to our Travis builds. flake8 and pylint are now separate builds, coverage is only reported on the py36 build.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,243 @@
+[MASTER]
+# Pickle collected data for later comparisons.
+persistent=no
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=migrations,config
+
+[MESSAGES CONTROL]
+# Disable the message, report, category or checker with the given id(s). You
+# can either give multiple identifier separated by comma (,) or put this option
+# multiple time (only on the command line, not in the configuration file where
+# it should appear only once).
+disable=missing-docstring,
+        too-few-public-methods,
+
+[REPORTS]
+# Set the output format. Available formats are text, parseable, colorized, msvs
+# (visual studio) and html
+output-format=colorized
+
+# Tells whether to display a full report or only the messages
+reports=yes
+
+# Activate the evaluation score.
+score=yes
+
+[LOGGING]
+# Logging modules to check that the string format arguments are in logging
+# function parameter format
+logging-modules=logging
+
+[FORMAT]
+# Maximum number of characters on a single line.
+max-line-length=80
+
+# Maximum number of lines in a module
+max-module-lines=1000
+
+# String used as indentation unit. This is usually " " (4 spaces) or "\t" (1
+# tab).
+indent-string='    '
+
+# List of optional constructs for which whitespace checking is disabled. `dict-
+# separator` is used to allow tabulation in dicts, etc.: {1  : 1,\n222: 2}.
+# `trailing-comma` allows a space between comma and closing bracket: (a, ).
+# `empty-line` allows space-only lines.
+no-space-check=trailing-comma,
+               dict-separator
+
+[TYPECHECK]
+# Tells whether missing members accessed in mixin class should be ignored. A
+# mixin class is detected if its name ends with "mixin" (case insensitive).
+ignore-mixin-members=yes
+
+# List of class names for which member attributes should not be checked (useful
+# for classes with dynamically set attributes). This supports the use of
+# qualified names.
+ignored-classes=optparse.Values,thread._local,_thread._local,Generator
+
+# List of members which are set dynamically and missed by pylint inference
+# system, and so shouldn't trigger E0201 when accessed. Python regular
+# expressions are accepted.
+generated-members=objects,DoesNotExist
+
+[MISCELLANEOUS]
+# List of note tags to take in consideration, separated by a comma.
+notes=FIXME,HACK,TODO
+
+[SIMILARITIES]
+# Minimum lines number of a similarity.
+min-similarity-lines=5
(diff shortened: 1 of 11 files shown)


.pylintrc
```diff
@@ -0,0 +1,243 @@
+[MASTER]
```
infinitewarp: How much of this `.pylintrc` file is "stock" versus customized? Assuming `.pylintrc` files just override the built-in defaults, I think we should trim this down only to project-specific overrides.

cloudigrade/account/views.py
```diff
@@ -7,6 +7,7 @@
 
 
 class AccountViewSet(mixins.CreateModelMixin, viewsets.ReadOnlyModelViewSet):
+    # pylint: disable=too-many-ancestors
```
infinitewarp: Given that using multiple mixins is kinda _by design_ with DRF, I'm inclined to throw this to the rc file instead of here, but I guess this is OK for now if it's the only instance.

cloudigrade/util/tests/test_aws.py
```diff
@@ -29,7 +29,7 @@ def test_extract_account_id_from_arn(self):
     def test_error_extract_account_id_from_invalid_arn(self):
         """Assert error in account ID extraction from a badly-formed ARN."""
         mock_arn = faker.Faker().text()
-        with self.assertRaises(Exception):  # TODO more specific exceptions
```
infinitewarp: Deleting a `TODO` without actually fixing it? That's a paddlin'. 🏏

cloudigrade/account/serializers.py
```diff
@@ -44,13 +44,14 @@ def create(self, validated_data):
 
 
 class ReportSerializer(serializers.Serializer):
+    # pylint: disable=abstract-method
```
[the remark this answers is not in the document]
ilyavw: The problem here is that we do not implement the `create()` and `list()` methods.

cloudigrade/account/views.py
ilyavw: So far this was the only occurrence, I think if it shows up more times and keeps being a false positive then yea, it belongs permanently disabled.

cloudigrade/util/tests/test_aws.py
ilyavw: I thought about it, but what does that `TODO` really accomplish? What exceptions do we want to check? With how boto3 handles them, is it even worth pursuing further?

.pylintrc
ilyavw: That's correct. This is already a fairly cut-down version, I've kept some things that I figured would be nice to expose in case we ever want to tweak them, but if you'd like I can trim this down just to the customized values.

cloudigrade/util/tests/test_aws.py
ilyavw: On second review, that test didn't actually have anything to do with boto3, so I went ahead and expanded it a tad.

.pylintrc
ilyavw: Went ahead and cut it down just to the customized options.

cloudigrade/util/tests/test_aws.py
```diff
@@ -28,9 +28,13 @@ def test_extract_account_id_from_arn(self):
 
     def test_error_extract_account_id_from_invalid_arn(self):
         """Assert error in account ID extraction from a badly-formed ARN."""
-        mock_arn = faker.Faker().text()
-        with self.assertRaises(Exception):  # TODO more specific exceptions
-            aws.extract_account_id_from_arn(mock_arn)
+        bad_index_arn = 'The quick brown fox jumps over the lazy dog.'
+        empty_arn = '::::'
+
+        with self.assertRaises(IndexError):
```
infinitewarp: The intent of the `TODO` was actually to implement *our own* more specific exception so we could intelligently catch and present situation-appropriate messaging later, e.g. `InvalidArnException`. I guess that can wait for later when/if we do more user-friendly error handling, though, since it hasn't been needed yet.

ilyavw: So that was my original concern with that `TODO`, it simply did not have enough context, hence I felt it was "useless". I'm more than happy to rollback the `todo` if you wish though.

.pylintrc
```diff
@@ -0,0 +1,67 @@
...
+ignored-classes=optparse.Values,thread._local,_thread._local,Generator
+
+# List of members which are set dynamically and missed by pylint inference
+# system, and so shouldn't trigger E0201 when accessed. Python regular
+# expressions are accepted.
+generated-members=objects,DoesNotExist
+
+[MISCELLANEOUS]
+# List of note tags to take in consideration, separated by a comma.
+notes=FIXME,HACK,TODO
```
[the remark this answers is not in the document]
ilyavw: Unfortunately any score less than 10 is a failure (and I haven't found a way to change that), therefore it doesn't really matter how much it's nudged, it'll fail it regardless.

cloudigrade/util/tests/test_aws.py
infinitewarp: Fair point. :) I was used to previous teams I'd worked with recognizing "more specific exceptions" as the shorthand cue for implementing our own, but you're right that I should have been more specific here with my intentions.

Yeah, let's put that `TODO` back as something like
> TODO implement our own situation-appropriate exception here, e.g. `InvalidArnException`

ilyavw:
After much discussion we've come to the conclusion that pylint simply isn't a good fit for us.

Issue #105, opened 2018-03-07
Refactor models to support other clouds


infinitewarp:
As a Doppler developer, I want models designed so that AWS isn't the only first-class cloud provider so that later I can support Azure, GCP, etc.

## Acceptance Criteria 

- [x] Verify that database tables have been migrated to new scheme.
- [x] Verify that HTTP APIs have been updated to reflect flexibility in cloud types.
  - [x] Verify that creating an account requires a "cloud provider" identifier as an input.
  - [x] Verify that the returned account JSON includes the cloud provider in each instance
- [ ] ~Verify that CLI commands have been updated to reflect flexibility in cloud types.~

## Tech assumptions

- We probably want some parent-child class relationship for our models. This could include all of the ones defined so far.
- Some "cloud provider" ID/value should be included in the account creation, but everything else going forward can continue to just use the IDs that reference back to other models.

infinitewarp:
demo: https://asciinema.org/a/170978

PR #119, opened 2018-03-13
refactor to support multiple cloud providers

infinitewarp:
https://github.com/cloudigrade/cloudigrade/issues/105

refactor many things to make more cloud-agnostic


cloudigrade/account/reports/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+    Returns:
+        dict: Results as a dict-of-dicts-of-lists as described above.
+    """
+    grouped_events = collections.defaultdict(
+        functools.partial(collections.defaultdict, list)
+    )
+
+    for event in events:
+        product_key = cloud_helper.get_event_product_identifier(event)
+        instance_key = cloud_helper.get_event_instance_identifier(event)
```
adberglund: Does the Django ORM populate the instance object when pulling the initial events or will this make another DB call for each event?

adberglund: Does the Django ORM populate the instance object when pulling the initial events or will this make another DB call for each event?

cloudigrade/account/serializers.py
```diff
@@ -5,61 +5,89 @@
...
+
+    The serialized response includes a special "detail" field that links
+    to the cloud provider specific account.
+    """
 
     class Meta:
         model = Account
-        fields = ('id', 'url', 'account_id', 'account_arn')
-        read_only_fields = ('account_id', )
+        fields = (
```
adberglund: Does this guarantee the order the fields will be serialized in? If so: I like alphabetical, but having the dates together can be beneficial.

cloudigrade/account/reports/helper.py
```diff
@@ -0,0 +1,115 @@
...
+        cloud_provider (str): The cloud provider
+        cloud_account_id (str): The cloud-specific account ID
+
+    Returns:
+        ReportHelper: An instance of a cloud-specific ReportHelper child class.
+
+    """
+    if cloud_provider not in CLOUD_PROVIDERS:
+        raise InvalidCloudProviderError(
+            _('Unsupported cloud provider "{0}".').format(cloud_provider)
```
ilyavw: are we not doing the `f'{cloud_provider}'` string format method anymore?

cloudigrade/account/reports/helper.py
```diff
@@ -0,0 +1,115 @@
...
+        Returns:
+            str: Some relatively unique string for grouping by instance
+
+        """
+
+
+class AwsReportHelper(ReportHelper):
+    """Report helper for AWS."""
+
+    def assert_account_exists(self):  # noqa: D102, see parent class
```
ilyavw: I don't think there is any harm in inserting an aws specific docstring for this method, better than littering it with `noqa` tags.

cloudigrade/account/serializers.py
infinitewarp: No, and I actually believe DRF always automatically alphabetizes the fields when outputting as JSON.

cloudigrade/account/reports/helper.py
infinitewarp: Yes, I mentioned this in a previous PR. I believe f-strings often don't work for gettext because of the order of execution. Which reminds me I need to file an issue to go fix/clean up all our incorrect f-strings…

Take this for example:
```
astronaut = 'Dave'
message = _(f"I'm sorry, {astronaut}.")
```
The value passed into the `_` function would have already been substituted by the Python interpreter, meaning the translation function gets `"I'm sorry, Dave."` which should not be found in the translation mapping.

cloudigrade/account/reports/helper.py
infinitewarp: Yeah, doing this felt iffy, and I was wondering how you guys would feel about it. On the other hand, copy-pasting large swaths of doc strings doesn't feel any better. I wish there was a better supported way to stay DRY with them. 😕

cloudigrade/account/reports/__init__.py
infinitewarp: Django's ORM is generally lazy/deferred in its execution, and it only executes queries when it thinks it needs to. In this case, the act of calling the iterator should have executed the query to get the events, and by including the `select_related` function in the chain, it *should* have pre-fetched the related data all at once.

See also:
https://docs.djangoproject.com/en/dev/ref/models/querysets/#select-related

infinitewarp: Also: https://stackoverflow.com/a/6388923

adberglund: Excellent!

cloudigrade/account/reports/helper.py
ilyavw: Maybe not, but in case we ever use anything to auto generate docs from the methods it would still be nice that there would be something written down.

PR #120, opened 2018-03-13
106 setup queue

adberglund:
Basic setup to start a RabbitMQ server in a Docker container

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -32,6 +32,16 @@ services:
       - "8000:80"
     depends_on:
       - app
-
+  queue:
+    image: rabbitmq:alpine
+    hostname: cloudiqueue
+    ports:
+      - "5672:5672"
+    environment:
+      - RABBITMQ_DEFAULT_USER=cloudigrade
+      - RABBITMQ_DEFAULT_PASS
+      - RABBITMQ_DISK_FREE_LIMIT={mem_relative, 1.5}
+    volumes:
+      - ./rabbitmq:/var/lib/rabbitmq
 volumes:
   static:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,8 +38,6 @@ services:
     ports:
       - "5672:5672"
     environment:
-      - RABBITMQ_DEFAULT_USER=cloudigrade
-      - RABBITMQ_DEFAULT_PASS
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -32,6 +32,16 @@ services:
...
       - app
-
+  queue:
+    image: rabbitmq:alpine
+    hostname: cloudiqueue
+    ports:
+      - "5672:5672"
+    environment:
+      - RABBITMQ_DEFAULT_USER=cloudigrade
+      - RABBITMQ_DEFAULT_PASS
```
ilyavw: This is something that is used locally, and only in dev, therefore should we go ahead and set a default? I see no harm in making this `RABBITMQ_DEFAULT_PASS=cloudigrade`.

requirements/base.txt
```diff
@@ -2,6 +2,7 @@ django==2.0.3
 django-environ==0.4.4
 boto3==1.6.7
 django-model-utils==3.1.1
+pika==0.11.2
```
ilyavw: I see we pull `pika` in, but we don't actually use it anywhere. Remove it?

adberglund: Sure. It will come into play in the next story, and I'll use it to demo, but it doesn't need to be here yet.

docker-compose.yml
adberglund: Sounds good, if we want to keep it in dev mode I can just remove the default user and pass vars and it will default to `guest:guest`

requirements/base.txt
infinitewarp: Mostly because I'm not familiar with the differences: why Pika instead of Celery? Or is it not an either/or thing?

docker-compose.yml
```diff
@@ -32,6 +32,14 @@ services:
       - "8000:80"
     depends_on:
       - app
-
+  queue:
+    image: rabbitmq:alpine
+    hostname: cloudiqueue
```
ilyavw: is there a reason this is necessary? Do we not just want to refer to this as `queue`?

adberglund: If we don't specify it gets randomly generated by Docker. RabbitMQ persists data by node/host, creating a directory based on that relationship. If we specify it, it will always load and store data from `rabbit@cloudiqueue`. We won't actually persist data in a reusable way if we don't specify it.

PR #134, opened 2018-03-19
use polymorphic framework to simplify urls

infinitewarp:


infinitewarp:
also, demo: https://asciinema.org/a/170978

PR #137, opened 2018-03-20
add :Z to volumes to fix selinux permissions

blentz:


ilyavw:
Hey Brett, you'll want to either branch off directly in `cloudigrade` and do the work, or if you want to keep your fork allow codecov, travis, and pyup access to your namespace so the CI can run.

ilyavw:
Also if you could provide just a brief description as to why this is necessary and what exactly it does would be great, thanks!

kdelee:
From: https://docs.docker.com/storage/bind-mounts/

> Configure the selinux label
If you use selinux you can add the z or Z options to modify the selinux label of the host file or directory being mounted into the container. This affects the file or directory on the host machine itself and can have consequences outside of the scope of Docker.

>The z option indicates that the bind mount content is shared among multiple containers.
>The Z option indicates that the bind mount content is private and unshared.
>Use extreme caution with these options. Bind-mounting a system directory such as /home or /usr with the Z option renders your host machine inoperable and you may need to relabel the host machine files by hand.

Also of interest:
https://github.com/moby/moby/issues/30934

In my experience when using the :z option for mounting directories to the quipucords container, it appears from my normal user perspective that the contents of the directory become owned by the user in the container -- which in the quipucords scenario is root. So if I mount ~/quipucords/data to the container, then I become unable to delete things in there without switching to root.

blentz:
@ilyavw Okay, I'll set up the access. As for the why question - On SELinux-enabled operating systems (e.g. Fedora, CentOS, RHEL), the bind mounts are mounted with the host system's file context. We need to have Docker relabel the volume with a context that works inside the container, otherwise we get "permission denied" errors when starting the entrypoint.sh script and when accessing various config files.

PR #138, opened 2018-03-20
add :Z to volumes to fix selinux permissions

blentz:


blentz:
see #137 for additional discussion.

PR #152, opened 2018-03-22
get makefile working for RH flavors

blentz:


Makefile
```diff
@@ -42,11 +58,25 @@ user-authenticate:
...
+# the app server starts up after the web proxy.
+#
+# Users should also set:
+# setsebool -P container_connect_any on
 start-compose:
-	docker-compose up --build -d
+	@if [ $(FLAVOR) == 'redhat' ]; then \
+		docker-compose up --build -d web db queue; \
+		$(SUDO) chmod 777 $(SOCKETDIR); \
+		$(SUDO) chcon -l s0 $(SOCKETDIR); \
```
kdelee: It appears to me just from reading this that the substative difference is that on the RH box you are testing this on, SELinux is enabled. This could be the case on more than just fedora or RHEL :man_shrugging: Cursory search digs up the following unverified quote:  "SELinux is not just available for Red Hat, but it is also available out-of-the-box for a number of other distributions including Fedora, Ubuntu, Debian, Hardened Gentoo and Yellow Dog Linux."

Could test with `getenforce` or `sestatus`

Makefile
```diff
@@ -42,11 +58,25 @@ user-authenticate:
...
+		docker-compose up --build -d app; \
+	else \
+		docker-compose up --build -d; \
+	fi
 
 stop-compose:
 	docker-compose down
+	$(SUDO) rm -rf $(TMPDIR)
```
infinitewarp: I don't think destroying directories makes sense in `stop-compose`. If anywhere, I think it should go in `clean`, but even then I don't like the idea of a make command destroying *all* of my `TMPDIR` since other processes might be putting stuff there. Maybe target the specific files you want gone instead.

infinitewarp: I don't think destroying directories makes sense in `stop-compose`. If anywhere, I think it should go in `clean`, but even then I don't like the idea of a make command destroying *all* of my `TMPDIR` since other processes might be putting stuff there. Maybe target the specific files you want gone instead.

Makefile
blentz: Yes, SELinux is available for more than just RH flavors. But, my scope isn't "all things running SELinux", my scope is "RH distros". Implementing support for non-RH distros is outside my scope.

Makefile
blentz: The problem I'm solving is that the start-compose command creates files in /tmp/cloudigrade, but nothing cleans them up. So, we're creating temporary files, but not cleaning them up.

I'm fine with putting the clean-up somewhere else. But I don't think it makes sense to leave temporary files laying about.

PR #153, opened 2018-03-22
88 enqueue new ami

adberglund:
During account creation, if we see a machine image we have not yet seen, we add it to a message queue. 

https://asciinema.org/a/ze5kKgOUlCySjqd5erkRbzaoo


cloudigrade/account/models.py
```diff
@@ -85,6 +107,18 @@ def __repr__(self):
         return f'<AwsInstance {self.ec2_instance_id}>'
 
 
+class AwsMachineImage(MachineImage):
+    """MachineImage model for an AWS EC2 instance."""
+
+    ec2_ami_id = models.CharField(
+        max_length=256,
```
blentz: AMI IDs are 17 characters, with a 4 character prefix ("ami-"), for a total of 21 characters. (e.g. "i-08165b429be87c52e")

max_length can be reduced to 32 or 64 without any issue.

cloudigrade/account/models.py
```diff
@@ -53,6 +53,28 @@ def product_identifier(self):
         """
 
 
+class MachineImage(BaseModel):
+    """Base Class for A cloud VM image."""
+
+    TYPE = model_utils.Choices(
+        'Windows',
+        'Linux',
+        'Unix-like'
```
blentz: Amazon distinguishes between a couple types of AMI for billing reasons - 

1) Windows Server (Various flavors are priced differently based on feature set)
2) Named Linux Distros with paid support (e.g. RHEL, SuSE, Ubuntu)
3) "Generic Linux/UNIX" 

I'm not sure if this is relevant to what Cloudigrade is doing, but I'm skeptical that "Unix-like" is going to be a useful part of the model or that a single "Linux" flag is not going to need to be differentiated at some point.

cloudigrade/account/models.py
adberglund: We can definitely reduce. There's a pre-existing `AwsInstance` model with the same field, and I was matching length and field type here.

cloudigrade/account/models.py
adberglund: A little background: this is really serving as a flag for Windows instances so that we don't go through the trouble of trying to inspect a Windows machine. 

I was actually expecting some input/changes here. I don't like what I have here, but figured let's crowdsource! It could just be a boolean flag for determining if it is Windows. Doesn't actually need string choices. 
  
What is *not* in the model at this time, but will be in a story or two is more information coming back from our image inspection process. I'm leaving the model expansion to that story. At that time we'll likely store results that indicate actual distribution.

cloudigrade/account/tests/test_util.py
```diff
@@ -0,0 +1,145 @@
...
+
+from django.test import TestCase
+
+from account import AWS_PROVIDER_STRING, util
+from account.models import AwsAccount, AwsMachineImage
+from util import aws
+from util.tests import helper as util_helper
+
+
+def mock_publish(body, exchange, routing_key, declare):
```
infinitewarp: Instead of creating a throwaway function, you could mock the `return_value` of a `Mock` object in your test.

cloudigrade/account/tests/test_util.py
```diff
@@ -0,0 +1,145 @@
...
+        expected = {ami_list[0]: True}
+
+        with patch.object(util, 'kombu') as mock_kombu:
+            mock_exchange = mock_kombu.Exchange.return_value
+            mock_queue = mock_kombu.Queue.return_value
+            mock_conn = mock_kombu.Connection.return_value
+            mock_with_conn = mock_conn.__enter__.return_value
+            mock_producer = mock_with_conn.Producer.return_value
+            mock_pub = mock_producer.publish
+            mock_pub.side_effect = mock_publish
```
infinitewarp: As mentioned on the previous comment, I think this could be:
```
mock_pub.return_value = True
```

cloudigrade/account/tests/test_util.py
```diff
@@ -0,0 +1,145 @@
...
+            self.assertEqual(ami.ec2_ami_id, ami_id)
+
+    def test_create_new_machine_images_with_windows_image(self):
+        """Test that new windows machine images are marked appropriately."""
+        aws_account_id = util_helper.generate_dummy_aws_account_id()
+        arn = util_helper.generate_dummy_arn(aws_account_id)
+        account = AwsAccount(account_arn=arn, aws_account_id=aws_account_id)
+        account.save()
+
+        region = f'region-{uuid.uuid4()}'
```
infinitewarp: I've been meaning to clean up some older tests that do the same thing, but to get us closer to somewhat more realistic mock data, I'd like to take all our goofy UUID-based fake regions and swap them out with `random.choice(helper.SOME_AWS_REGIONS)`.

If you're feeling ambitious, you could do that here and in your other new tests… or just punt this for later. I have no strong opinion at the moment.

infinitewarp: Oh, I see you're actually using `random.choice` in some of the later tests! In that case, yeah, let's clean this up for consistency.

cloudigrade/account/util.py
```diff
@@ -45,3 +49,111 @@ def create_initial_aws_instance_events(account, instances_data):
...
+        )
+        new_amis = list(seen_amis.difference(known_amis))
+
+        for new_ami in new_amis:
+            ami = AwsMachineImage(
+                account=account,
+                ec2_ami_id=new_ami,
+                platform=(MachineImage.TYPE.Windows
+                          if platforms.get(new_ami)
+                          else MachineImage.TYPE.Linux)
```
infinitewarp: I feel lukewarm about explicitly setting "Linux" when we only really know if it's Windows or not-Windows at this point.

Maybe we need an "Unknown" state or to let it be nullable until we know more information.

Admittedly, it's *very* unlikely to be neither Windows or Linux, but the possibility exists.

cloudigrade/account/tests/test_util.py
adberglund: Oops. Will do.

cloudigrade/account/models.py
infinitewarp: 👍 to @adberglund's comment. We have some wonky oversized fields at present, and eventually we should go tighten them up, but I think this is fine right now for sake of consistency. The big field cleanup can come later.

cloudigrade/account/tests/test_util.py
adberglund: Yep, makes total sense. I was confusing myself into thinking that in order to use assert_called_with I needed a function, but realize that the mock will still get called with those now.

cloudigrade/account/util.py
adberglund: That's actually why I ended up making the model use a choice with Unix-like (what if they're running FreeBSD). But it's really about distinguishing windows or not windows. Would you prefer a boolean flag there instead of a choice?

infinitewarp: The more I think about it, the more I think a simple is-Windows boolean makes sense. That would cover exactly the case we currently need to know about (i.e. skipping AMIs for Windows), doesn't introduce ambiguity, and is easy to migrate/convert to a richer "choice" field later if we see value in it.

PR #158, opened 2018-03-26
Fix f-string translations

### LINKED ISSUE #130: Fix incorrect f-string translations
In our rush to start this project, I believe we made some incorrect assumptions about new-style format strings, and we didn't think about how they would be passed through translation functions.

Take this example:

```python
astronaut = 'Dave'
message = _(f"I'm sorry, {astronaut}.")
```

The value passed into the `_` function should have already been substituted by the Python interpreter, meaning the translation function gets "I'm sorry, Dave." which should not be found in the translation mapping.

So, the above example should be rewritten like the following:

```python
astronaut = 'Dave'
message = _("I'm sorry, {0}.").format(astronaut)
# or equally valid:
message = _("I'm sorry, {astronaut}.").format(astronaut=astronaut)
```

To verify that this fix is complete unfortunately requires someone to comb through the code to find all uses of the `_` translation function and verify that each is not being passed a f-string.

aweiteka:
Fixes #130 

Signed-off-by: Aaron Weitekamp <aweiteka@redhat.com>


infinitewarp:
Thanks for the contribution, @aweiteka! 🎉 

Unfortunately your change failed the build at Travis CI. Details:
https://travis-ci.org/cloudigrade/cloudigrade/jobs/358231405

Please take a look at our [CONTRIBUTING.rst](https://github.com/cloudigrade/cloudigrade/blob/master/CONTRIBUTING.rst) if you want info about running tests locally.

Issue #163, opened 2018-03-27
Convert base images to CentOS


infinitewarp:
As a Red Hat employee, I want cloudigrade images to be based on RHEL or a close derivative (e.g. CentOS) so that we have better in-house support.

## Acceptance Criteria 

- [x] Verify that all container images related to cloudigrade are built on some version of a RHEL-like distro

## Tech assumptions

- Is CentOS okay since it's downstream from RHEL? Or do we want Fedora since it's upstream of RHEL?

kdelee:
I prefer Fedora.
@elyezer what do you think?

elyezer:
@kdelee I don't have a strong opinion here. Though I think CentOS looks more server hardened than Fedora because it is considered the RHEL upstream.

But for an upstream project using Fedora is OK and offers more packages on its repository out of the box.

Maybe we can try to be consistent with other upstream projects and try to pick what they are picking, for example, quipucords is using Fedora for its Docker image.

ilyavw:
Just for fun, here are the old and new image sizes for the cloudigrade and houndigrade containers:
```
REPOSITORY                  TAG                 IMAGE ID            CREATED              SIZE
cloudigrade/houndigrade     alpine              fee90374d0a6        About a minute ago   138MB
cloudigrade/houndigrade     centos              24c652bad113        4 seconds ago        475MB
cloudigrade/cloudigrade     alpine              ef82def065b7        2 minutes ago        186MB
cloudigrade/cloudigrade     centos              2aea66f768bb        4 minutes ago        611MB
```

PR #168, opened 2018-03-28
Update container to be centos based

ilyavw:
#163


ilyavw:
@infinitewarp I've added size information for the cloudigrade and houndigrade containers to the story https://github.com/cloudigrade/cloudigrade/issues/163#issuecomment-377042447

PR #178, opened 2018-04-02
Modified EC2 snapshot API calls to not fail on undoc'd AWS validation

adberglund:
This will get us running again.

### DIFF
--- a/cloudigrade/util/aws.py
+++ b/cloudigrade/util/aws.py
@@ -285,17 +285,17 @@ def _verify_policy_action(session, action):
         elif action == 'ec2:DescribeSnapshotAttribute':
             ec2.describe_snapshot_attribute(
                 DryRun=True,
-                SnapshotId='string',
+                SnapshotId='snap-0f423c31dd96866b2',
                 Attribute='productCodes'
             )
         elif action == 'ec2:DescribeSnapshots':
             ec2.describe_snapshots(DryRun=True)
         elif action == 'ec2:ModifySnapshotAttribute':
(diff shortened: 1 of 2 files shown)


cloudigrade/util/aws.py
```diff
@@ -285,17 +285,17 @@ def _verify_policy_action(session, action):
         elif action == 'ec2:DescribeSnapshotAttribute':
             ec2.describe_snapshot_attribute(
                 DryRun=True,
-                SnapshotId='string',
+                SnapshotId='snap-0f423c31dd96866b2',
```
infinitewarp: Is this hard-coded ID intentional? Where is the magic that originally defined it?

infinitewarp: (Although I did see the conversation in Slack, we should at least comment in here why we have a magic token so that the rationale isn't lost to time.)

infinitewarp: And since we're repeating the magic token in several places (both here and in the tests) maybe it should get pulled into a constant-like value at the top of the module.

adberglund: The magic was originally the string `string`. The pattern _appears_ to be `snap-17hexdigits` but between Ilya and myself we have found 1 code that fails validation and 1 that does not, so I am not comfortable random generating it until AWS has sorted this out.

adberglund: And since they pulled the rug out beneath us anyway, this fix could start failing at any moment given that this change was not documented.

PR #179, opened 2018-04-02
generate a 400 error, instead of 500 when receiving an invalid ARN

blentz:


cloudigrade/util/aws.py
```diff
@@ -8,6 +8,7 @@
 import boto3
 from botocore.exceptions import ClientError
 from django.conf import settings
+from django.core.exceptions import SuspiciousOperation
```
ilyavw: Not sure if `SuspiciousOperation` really feels right here, even if we're just re-using it for the easy `400` we may inadvertently accidentally catch `InvalidArn` exceptions if we're catching `SuspiciousOperation` in the future when trying to catch actual security issues. I'd encourage you to extend [DRF `ApiException`](http://www.django-rest-framework.org/api-guide/exceptions/#api-reference) instead, maybe [ParseError](http://www.django-rest-framework.org/api-guide/exceptions/#parseerror).

cloudigrade/util/aws.py
```diff
@@ -70,6 +71,12 @@ def is_running(cls, code):
         return code == cls.running.value
 
 
+class InvalidArn(SuspiciousOperation):
```
ilyavw: This exception is probably better off living in `exceptions.py`.

blentz: I only see account/exceptions.py, which doesn't seem right. Are you meaning "create a new exceptions.py file" ?

ilyavw: Yup!

Issue #146, opened 2018-03-21
Write an architectural overview doc


infinitewarp:
As Dan, I want an architectural overview document so that I can show other stakeholders and potential customers how we will interact with their cloud to measure and report on their usage data.

This will allow us to take a customer who wants to report public cloud stuff and give them an idea of how we will be measuring it. This gets us early feedback in case a customer says "whoa we can't do that".

## Acceptance Criteria 

- [x] Verify document exists that includes:
  - [x] Description of interactions between system components
  - [x] Description of interactions between cloudigrade and a customer's AWS account
  - [x] Visualization of system components with technology labels

## Tech assumptions

- Talk to Dan as you are writing this for sanity-checking.
- Also talk to Dan about maybe producing visualizations/diagrams/pictures.

PR #185, opened 2018-04-04
add initial architecture overview docs

infinitewarp:
https://github.com/cloudigrade/cloudigrade/issues/146

Here's the first pass at an architectural overview document. This should not be considered exhaustively complete because I expect this to be a living document that we update as we progress through cloudigrade's development. Also, much of this document is _speculative_ since we haven't actually completed coding the workflow around inspection.

Here's a direct link to the rendered document:
https://github.com/cloudigrade/cloudigrade/blob/146-arch-docs/docs/architecture.rst


infinitewarp:
@ilyavw Good idea! I've added support for `make docs-seqdiag` in the latest commit.

Makefile
```diff
@@ -16,6 +16,7 @@ help:
 	@echo "  start-compose            to compose all containers in detached state"
 	@echo "  stop-compose             to stop all containers"
 	@echo "  start-db                 to start the psql db in detached state"
+	@echo "  docs-seqdiag             to regenerate docs .svg files from .diag files"
```
ilyavw: Would you mind making an overarching `make docs` that would call `docs-seqdiag` as part of it? That way we have a simple command to call that would generate _all_ docs (as soon as we have them 😄)

infinitewarp: Now we have `made doc` and `make doc-seqdiag`.

For context, we couldn't call it `make docs` because make doesn't like having a target be the same name as a local directory name. 🤦‍♂️ https://stackoverflow.com/a/4435491/6621995

Makefile
```diff
@@ -53,3 +55,9 @@ start-db:
 
 start-queue:
 	docker-compose up -d queue
+
+doc-seqdiag:
+	cd docs && for FILE in *.diag; do seqdiag -Tsvg $$FILE; done
+
+doc:
```
ilyavw: Proper syntax here would actually be `doc: doc-seqdiag`

dozdowski:
I really like the organization of this documentation, and I feel like the charts are an excellent developer-level overview of what happens as Cloudigrade operates.

I feel that two improvements are probably called for:

- We could work the grammar a little so paragraphs flow like composition rather than stereo instructions.  :) https://www.youtube.com/watch?v=RZBaSUhrwBY
- I think one extra image that presents a high level overview of the workflows represented in the task-driven diagrams would be awesome.  I'm willing to contribute that.

What's the best way to (A) kick around the wording and (B) contribute the overview image?

infinitewarp:
@dozdowski If you want to put on your developer hat, you're welcome to create another branch off of mine, make some edits, and submit that up as a pull request. 😉  Although, it's probably easier to throw some specific _inline_ suggestions at me here, and I'd be happy to make the edits. In the latter case, on the [Files](https://github.com/cloudigrade/cloudigrade/pull/185/files) link at the top of this PR, you can make inline comments on the individual documents by clicking the `[+]` that appears in the left margins between the line numbers and the contents. Scroll down and look for `docs/architecture.rst`.

As for extra diagrams, I can try to throw one together and get back to you to see if it fits the bill. Otherwise if you already have something in mind or already sketched out, it might be easier to just email or slack a picture to me.

dozdowski:
I went back to offer suggestions, and it looks like some cleanup was done.  As is, I think it reads very well.  I've created an "overview" sketch that I shot to Brad via PM.  I'm good with everything I see here, and we can add the overview image when appropriate.

PR #198, opened 2018-04-09
167 copy snapshot task

adberglund:
Hoh boy.


cloudigrade/account/tasks.py
```diff
@@ -0,0 +1,46 @@
...
+                             AwsSnapshotNotOwnedError)
+
+
+@shared_task(autoretry_for=(AwsSnapshotCopyLimitError,
+                            AwsSnapshotNotOwnedError),
+             retry_backoff=True,
+             max_retries=10)
+def copy_ami_snapshot(arn, ami_id, source_region):
+    """
+    Copy an AWS Snapshot to the primary AWS accout.
```
blentz: "account"

cloudigrade/account/tests/test_tasks.py
```diff
@@ -0,0 +1,128 @@
...
+
+        copy_ami_snapshot(mock_arn, mock_image_id, mock_region)
+
+        mock_aws.get_session.asssert_called_with(mock_arn)
+        mock_aws.get_ami.asssert_called_with(
+            mock_session,
+            mock_image_id,
+            mock_region
+        )
+        mock_aws.get_ami_snapshot_id.asssert_called_with(mock_image)
```
blentz: Should that be "assert_called_with" ?

cloudigrade/util/aws.py
```diff
@@ -220,6 +229,120 @@ def get_ec2_instance(session, instance_id):
...
+        OperationType=operation,
+        UserIds=[user_id]
+    )
+
+    # The modify call returns None. This is a check to make sure
+    # permissions are added successfully.
+    response = snapshot.describe_attribute(Attribute='createVolumePermission')
+
+    for user in response['CreateVolumePermissions']:
+        if user['UserId'] == user_id:
```
blentz: I don't think this will work in the "remove" case. Removing a permission, by definition, means there will be no match and, therefore, this will fall through to raising an exception. Given that "remove" is a valid option to pass in, and your docstring doesn't mention raising an exception as valid/expected behavior, I think this is a bug.

IMO, I think you'd be better off refactoring this into two methods: add_owner_to_snapshot() and remove_owner_from_snapshot().

cloudigrade/config/settings/base.py
```diff
@@ -155,4 +155,4 @@
 
 RABBITMQ_EXCHANGE_NAME = 'cloudigrade_inspectigrade'
 RABBITMQ_QUEUE_NAME = 'machine_images'
-RABBITMQ_URL = 'amqp://guest:guest@localhost:5672/%2F'
+CELERY_BROKER_URL = 'amqp://guest:guest@localhost:5672/%2F'
```
infinitewarp: Out of curiosity, did you ever encounter a need to set `CELERY_RESULT_BACKEND`? It might not be necessary since we aren't checking for results from calling a task, but when I was experimenting locally, I *think* I had to set `CELERY_RESULT_BACKEND` to the same value as `CELERY_BROKER_URL` to make that work.

cloudigrade/util/tests/helper.py
```diff
@@ -155,6 +155,58 @@ def generate_mock_ec2_instance(instance_id=None, image_id=None, subnet_id=None,
     return mock_instance
 
 
+def generate_mock_image(image_id=None, encrypted=False):
```
blentz: Does it make sense to use Moto ( https://github.com/spulec/moto ) instead of mocking out AWS objects by hand?

cloudigrade/util/exceptions.py
```diff
@@ -6,3 +6,27 @@ class InvalidArn(ParseError):
     """an invalid ARN was detected."""
 
     pass
+
+
+class AwsValueError(ValueError):
+    """Inapropriate value given to function."""
+
```
infinitewarp: You can delete the `pass`es from these exception classes. Having a docstring like this is good enough to keep the structure in place.

cloudigrade/util/aws.py
```diff
@@ -220,6 +229,120 @@ def get_ec2_instance(session, instance_id):
...
+        raise AwsValueError(_('{o} not one of {ao}').format(
+            o=operation,
+            ao=acceptable_operations)
+        )
+
+    snapshot = session.resource('ec2',
+                                region_name=region).Snapshot(snapshot_id)
+
+    if snapshot.encrypted:
+        logger.warning(_('Snapshot is encrypted. Not copying.'))
```
ilyavw: This operation is to add our account to owners list for the snapshot, is the snapshot being encrypted relevant in this case? Either way it doesn't look like any copying actually happens here, we should probably update the error.

cloudigrade/util/tests/helper.py
```diff
@@ -155,6 +155,58 @@ def generate_mock_ec2_instance(instance_id=None, image_id=None, subnet_id=None,
...
+
+    mock_image = Mock()
+    mock_image.image_id = image_id
+    mock_image.root_device_name = root_device_name
+    mock_image.root_device_type = root_device_type
+    mock_image.block_device_mappings = block_device_mappings
+    return mock_image
+
+
+def generate_mock_snapshot_id():
```
infinitewarp: I did something very similar in my branch! 😆  FWIW, my implementation was:

```
def generate_dummy_snapshot_id():
    """Generate a dummy AWS snapshot ID for testing purposes."""
    return 'snap-{}'.format(
        ''.join(random.choice(string.hexdigits[:16]) for _ in range(17))
    )
```
Although it's not strictly necessary, for more realistic random values, you should slice `[:16]` like I did or else it's twice as likely to get values a-f because `hexdigits` includes both lower and upper case letters.

cloudigrade/config/settings/base.py
adberglund: I have not yet. Since our current design chains tasks and uses message queues as go-betweens, I didn't see a need to track the result directly here. I believe Celery is smart enough to ignore results until you give it a back end.

cloudigrade/util/aws.py
```diff
@@ -220,6 +229,120 @@ def get_ec2_instance(session, instance_id):
...
+        raise AwsSnapshotEncryptedError
+
+    attribute = 'createVolumePermission'
```
ilyavw: Do we actually need `createVolumePermission`? It is my understanding that we only need to add our `userId` to the snapshot, not the `createVolumePermission` list too. Since we're copying the snapshot, not creating the volume from it. We create the volume from _our_ copied version.

cloudigrade/account/tests/test_tasks.py
adberglund: Yes. Thanks mocking for letting that fly... ☠️

cloudigrade/util/tests/helper.py
infinitewarp: Yeah, on second though, let's just leave yours and I'll twiddle the merge in place when mine branch arrives. 👍

cloudigrade/util/aws.py
adberglund: I'd love to use the minimum API call that works here. The best I can tell we have to use modify_attribute, and that has two options for attribute. either `productCodes` or `createVolumePermission` and we've been encountering funky validation with this specific API call recently. As written, it was the only way I could get the snapshot to be modified in a way that we could copy it.

If I'm missing something, please add here.

ilyavw: @adberglund you're right, confusion on my part with the docs.

cloudigrade/account/tasks.py
```diff
@@ -0,0 +1,46 @@
...
+        source_region (str): The region the snapshot resides in
+
+    Returns:
+        None: Run as an asyncronous Celery task.
+
+    """
+    session = aws.get_session(arn)
+    ami = aws.get_ami(session, ami_id, source_region)
+
+    try:
```
adberglund: This is actually a mess right now, fixing with the refactor of snapshot ownership functions in aws.

cloudigrade/config/settings/base.py
```diff
@@ -155,4 +155,8 @@
 
 RABBITMQ_EXCHANGE_NAME = 'cloudigrade_inspectigrade'
```
infinitewarp: Be sure to rebase/merge from master to get @ilyavw's recent change to this

cloudigrade/config/settings/base.py
infinitewarp: Related to the previous comment about including the latest change from master, this config's value should also be pulled via `env`.

adberglund:
https://asciinema.org/a/obJLfd5vtYVjQPH9pxutcFGvN

Issue #149, opened 2018-03-21
Test environment


infinitewarp:
As a quality engineer, I want a test environment so that I can run integration tests.

## Acceptance Criteria 

- [x] Verify we have cloudigrade, postgresql, and rabbitmq running in OpenShift
- [x] Verify that new releases are automatically being deployed to the environment
- [x] Verify that database migrations are automatically running during deployment
- [ ] Verify that after an automatic release is deployed to the test environment, the test suite is triggered to run.
- [x] Verify that project documentation has been update to explain how this works.

## Tech assumptions

- This includes the initial setup of OpenShift with cloudigrade.
- The "how our updates get installed" question is to be experimented and answered here. 
- Remember that we need to deal with templating, secret management, etc. here.
- Ideally whoever is working on this will also teach the group about how things work. Lunch and learn, perhaps?
- Let's try doing proper continuous development and have each commit on master trigger a release to the test environment.
- As we are building this, have continued conversations about including a hook to initiate the actual running of tests.
- Decide whether to reuse or create a new cluster in AWS for houndigrade.

aweiteka:
It would be slick to use OpenShift source-to-image builds for this. Documentation suggests OpenShift supports up to [python 3.5](https://docs.openshift.org/latest/using_images/s2i_images/python.html#s2i-images-python-versions), although I see 3.6 support in the [S2I repo](https://github.com/sclorg/s2i-python-container/tree/master/3.6). This project uses at least one [3.6 feature](https://docs.python.org/3/whatsnew/3.6.html#pep-498-formatted-string-literals).

Definitely worth checking into. +1

PR #206, opened 2018-04-11
Tweak our docker configuration

ilyavw:
Tweaked several things in our docker config while making changes for #149 
* Remove user from nginx.conf
* Set nginx port to 8080
* **Update compose to listen and serve cloudigrade on 8080**
* Update compose to no longer use a named volume for static file
* Update compose to no longer mount the entrypoint.sh in the cloudigrade directory
* Update entrypoint script to collect static files
* Update dockerfile to no longer package staticfiles in the container
* Update nginx/compose/cloudigrade docker settings file indicating new static file path

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,7 +9,7 @@ services:
       - ./pg_data:/var/lib/postgresql
   app:
     build: .
-    entrypoint: /opt/cloudigrade/entrypoint.sh
+    entrypoint: /opt/entrypoint.sh
     environment:
       - DJANGO_SETTINGS_MODULE=config.settings.docker
       - AWS_ACCESS_KEY_ID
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -9,27 +9,27 @@ services:
...
+      - /tmp/cloudigrade/static:/srv/cloudigrade/static:Z
     depends_on:
       - db
   web:
     image: nginx:alpine
     volumes:
       - ./docker/nginx.conf:/etc/nginx/nginx.conf:Z
       - /tmp/cloudigrade/socket:/var/run/cloudigrade:Z
-      - static:/opt/cloudigrade/static
+      - /tmp/cloudigrade/static:/srv/cloudigrade/static:Z
```
infinitewarp: Does something need to `mkdir` this first?

ilyavw: Compose handles it.

ilyavw:
@infinitewarp containers can't run as root in openshift, so they can't bind to port 80. I'd like to reuse the nginx config out of this directory(and generate s2i nginx image from here too), therefore it makes sense to have a single nginx.conf across the board.

We are dropping the static _named_ volume due to it being mounted as root inside the container, which doesn't actually allow us to place any files in it at container run-time.

PR #209, opened 2018-04-11
Make our settings more consistent

ilyavw:


cloudigrade/config/settings/base.py
```diff
@@ -94,7 +94,7 @@
 DATABASES = {
     'default': env.db('DATABASE_URL', default='postgres://localhost/cloudigrade'),
 }
-DATABASES['default']['ATOMIC_REQUESTS'] = True
+DATABASES['default']['ATOMIC_REQUESTS'] = env('DJANGO_ATOMIC_REQUESTS', default=True)
```
infinitewarp: Do you foresee us ever wanting to override this at the project-environment level?

cloudigrade/config/settings/prod.py
```diff
@@ -1,9 +1,11 @@
 from .base import *  # noqa
 
+DEBUG = env.bool('DJANGO_DEBUG', default=False)
```
infinitewarp: I believe this should be `False`. When would we want to enable this in prod?

If we do think this could happen, then we should drop the `DEBUG = ...` from each of the environment-specific files and just let it be in base.

cloudigrade/config/settings/prod.py
```diff
@@ -1,9 +1,11 @@
...
 SECRET_KEY = env('DJANGO_SECRET_KEY')
-ALLOWED_HOSTS = env('DJANGO_ALLOWED_HOSTS')
+ALLOWED_HOSTS = env.list('DJANGO_ALLOWED_HOSTS')
 DATABASES = {
     'default': {
+        'ATOMIC_REQUESTS': env('DJANGO_ATOMIC_REQUESTS', default=True),
```
infinitewarp: Same as before about this…

cloudigrade/config/settings/base.py
infinitewarp: When would we ever override this?

ilyavw:
@infinitewarp I've made some changes based on our conversation. I've replaced `env.db()` in most places to simplify configuration, broken up the `RABBITMQ_URL` to make it easier to update, and added comments to the config files clarifying their purpose.

Finally for the sake of debugging I did update the rabbitmq compose image to run the version that includes the management webapp.

PR #216, opened 2018-04-12
WIP update README for fedora users

kdelee:
This is a work in progress, but I thought I'd go ahead push it in case other Fedora users get started and to ask some questions.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -8,7 +8,12 @@ cloudigrade
 What is cloudigrade?
 ====================
 
-**cloudigrade** is an open-source suite of tools for tracking Linux distribution use (although chiefly targeting RHEL) in public cloud platforms. **cloudigrade** actively checks a user's account in a particular cloud for running instances, tracks when instances are powered on, determines what Linux distributions are installed on them, and provides the ability to generate reports to see how long different distributions have run in a given window.
+**cloudigrade** is an open-source suite of tools for tracking Linux
+distribution use (although chiefly targeting RHEL) in public cloud platforms.
+**cloudigrade** actively checks a user's account in a particular cloud for
+running instances, tracks when instances are powered on, determines what Linux
+distributions are installed on them, and provides the ability to generate
+reports to see how long different distributions have run in a given window.
 
 
 What is this "Doppler" I see referenced in various places?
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -82,20 +99,43 @@ Once you have an environment set up, install our Python package requirements:
...
+The following installation method has been tested on Fedora 26. Some packages
+must be installed on the system level and are available via ``dnf``, the python
+specific dependencies can be installed into a virtual environment.
+
+.. code-block:: bash
+
+    sudo dnf install gettext postgresql postgresql-devel python3 libffi-devel libffi
+    python3 -m venv ~/envs/cloudigrade
+    source ~/envs/cloudigrade/activate
+    pip install -U pip awscli wheel tox
```
kdelee: Is there a good reason why we can't just include awscli wheel and tox in the requirements or in some setup.py?

README.rst
```diff
@@ -191,13 +238,21 @@ This auth token can be supplied in the Authorization header.
...
 
-RabbitMQ is used to broker messages between **cloudigrade** and inspectigrade services. There are multiple Python packages available to interact with RabbitMQ; the officially recommended packaged is `Pika <https://pika.readthedocs.io/en/latest/>`_. Both services serve as producers and consumers of the message queue. The **cloudigrade** docker-compose file requires that a password environment variable be set for the RabbitMQ user. Make sure that the following has been set in your local environment before starting
+RabbitMQ is used to broker messages between **cloudigrade** and inspectigrade
+services. There are multiple Python packages available to interact with
+RabbitMQ; the officially recommended packaged is `Pika
+<https://pika.readthedocs.io/en/latest/>`_. Both services serve as producers
+and consumers of the message queue. The **cloudigrade** docker-compose file
+requires that a password environment variable be set for the RabbitMQ user.
+Make sure that the following has been set in your local environment before
+starting
```
kdelee: I did not see any information before this section about installing or setting up RabbitMQ. 
Does there need to be a section about that? Does it allready exist somewhere else?

infinitewarp: Since we only very recently started actively using RabbitMQ, we may need to expand this. I'm taking a note to do that.

FWIW, as long as you're using docker-compose, you don't need to install anything locally. We just pull a public container image and use that.

kdelee:
Going to close this for now since there are changes afoot

Issue #176, opened 2018-04-02
Run scheduled task to scale up cluster size


infinitewarp:
As a Doppler admin, I want the cluster size to automatically scale up via a *scheduled* Celery task (like cron) so the cluster becomes available automatically when some volumes are ready.

## Acceptance Criteria 

- [x] Verify that when the cluster size is nonzero, this process exits cleanly with no changes.
- [x] Verify that when the cluster size is zero and there are zero messages, this process exits cleanly with no changes.
- [x] Verify that when the cluster size is zero and there are some messages representing ready volumes, this process dequeues up to the configured limit number of messages and sets the cluster size to 1.
- [x] Verify that when there are *more* messages on the queue than the batch size limit, this process reads and dequeues only up to the configured limit number of messages.

## Tech assumptions

- If the scale is already nonzero, then we must exit because this means the cluster is currently active.
- We do *not* need to retry later in the first acceptance criteria as long as we have not yet dequeued/lost any messages and because the schedule should automatically start again later.
- Because AWS limits the number of volumes we can mount, we need a limit baked into the part of the process. This should be a configuration variable with a reasonable default (32).
- The estimate we have given to this story assumes that _something_ exists within the boto3 library to do what we need (we have not yet verified this).

ilyavw:
This looks like the boto3 api we'll need to use http://boto3.readthedocs.io/en/latest/reference/services/application-autoscaling.html#ApplicationAutoScaling.Client.put_scaling_policy

infinitewarp:
- demo for AC1: https://asciinema.org/a/178750
- demo for AC2: https://asciinema.org/a/178752
- demo for AC3: https://asciinema.org/a/178755

I would ask that the "demo" for AC4 be to review and trust the _thorough_ unit test coverage. 🙂

ilyavw:
> I would ask that the "demo" for AC4 be to review and trust the thorough unit test coverage. 🙂

404 Test Not Found

infinitewarp:
I added another unit test, and I recorded another demo!

- demo for AC4: https://asciinema.org/a/178767

ilyavw:
AC Verified

PR #224, opened 2018-04-20
Organize AWS helpers

infinitewarp:
Recently @ilyavw suggested we break up the `aws` module because it was starting to get a bit long. On the surface this seemed like a reasonable request, but actually tearing things apart without causing dependency cycles and preserving passing tests became a small nightmare. Fortunately, I think I've slogged through it all and come out victorious.

In the process, though, I had to look at every one of our aws helper tests and modify *most* of them in some way. As I was moving and modifying them, I realized that some of them deserved a deeper cleanup (read: "a rewrite") and I figured that now was a good time to do that. Thus, this massive change doesn't just lift and move code around; now *some of* the tests are totally different.

I would kindly ask you as a reviewer to pay closer attention as you skim through the tests. Many of them should look very familiar since they are practically copy-paste jobs, but some of them are brand new and warrant a more critical eye.

The commits here were originally part of a new feature branch for #176, but since I found myself adding to the `aws` module, I naively thought I should do this refactoring first. 🙂 The PR for the actual functionality requested in #176 will come later.


cloudigrade/util/aws/ec2.py
```diff
@@ -0,0 +1,257 @@
...
+import enum
+import logging
+
+import boto3
+from botocore.exceptions import ClientError
+from django.utils.translation import gettext as _
+
+from util.aws.helper import get_regions
+from util.aws.sts import _get_primary_account_id
+from util.exceptions import (AwsSnapshotCopyLimitError,
```
ilyavw: This seems to be a different way to import multiple objects vs the way its done in `aws/__init__.py`. Is there a reason? Should we try and make this more consistent?

cloudigrade/util/tests/test_aws_arn.py
```diff
@@ -0,0 +1,95 @@
+"""Collection of tests for ``util.aws.helper`` module."""
```
ilyavw: `util.aws.helper` -> `util.aws.arn`?

cloudigrade/util/aws/ec2.py
infinitewarp: @ilyavw Yep, this was PyCharm misbehaving when I carved up all these files. Should be fixed in various files with the latest commit!

PR #229, opened 2018-04-25
scheduled task to conditionally scale up the houndigrade cluster

infinitewarp:
https://github.com/cloudigrade/cloudigrade/issues/176

This PR adds our first support for scheduled Celery tasks using beat. As described in https://github.com/cloudigrade/cloudigrade/issues/176, this task checks the current cluster auto scaling size, checks the message queue for volume IDs, scales up to 1, and runs a (currently stubbed-out) async task that will be responsible for actually "running" houndigrade in the cluster.

Note: this PR builds upon the `organize-aws-helpers` branch and needs to merge _after_ #224 merges.

Demo: coming soon!

### DIFF
--- a/cloudigrade/account/tasks.py
+++ b/cloudigrade/account/tasks.py
@@ -1,7 +1,10 @@
 """Celery tasks for use in the account app."""
+import logging
+
 from botocore.exceptions import ClientError
 from celery import shared_task
 from django.conf import settings
+from django.utils.translation import gettext as _
 
 from account.models import AwsMachineImage
 from account.util import add_messages_to_queue, read_messages_from_queue
@@ -9,6 +12,8 @@ from util import aws
 from util.aws import rewrap_aws_errors
 from util.celery import retriable_shared_task
 from util.exceptions import AwsSnapshotEncryptedError
+
+logger = logging.getLogger(__name__)
 
 
 @retriable_shared_task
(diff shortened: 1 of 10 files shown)


cloudigrade/config/settings/base.py
```diff
@@ -30,7 +30,16 @@
...
-                                        default='us-east-1a')
+                                        default='us-east-1b')
+HOUNDIGRADE_AWS_AUTOSCALING_GROUP_NAME = env(
+    'HOUNDIGRADE_AWS_AUTOSCALING_GROUP_NAME',
+    default='EC2ContainerService-inspectigrade-test-bws-us-east-1b-'
+            'EcsInstanceAsg-JG9NX9WHX6NU'
+)
+HOUNDIGRADE_AWS_VOLUME_BATCH_SIZE = env.int(
+    'HOUNDIGRADE_AWS_VOLUME_BATCH_SIZE',
+    default=5
```
ilyavw: Do we not want to do 32 like we originally talked about?

infinitewarp: Good catch. 👍  I accidentally overlooked that requirement.

cloudigrade/account/tasks.py
```diff
@@ -85,4 +87,52 @@ def enqueue_ready_volume(ami_id, volume_id, region):
...
+        # Quietly exit and let a future run check the scaling.
+        return
+
+    messages = read_messages_from_queue(
+        'ready_volumes',
+        settings.HOUNDIGRADE_AWS_VOLUME_BATCH_SIZE
+    )
+
+    if len(messages) == 0:
+        # Quietly exit and let a future run check for messages.
```
ilyavw: Not sure if this is the right time, but I'd love to see some debug logging happen in situations like this.

infinitewarp: I'd considered about the same thing when I wrote this. I think I'll add some in now.

cloudigrade/config/settings/base.py
```diff
@@ -190,3 +200,9 @@
     'account.tasks.create_volume': {'queue': 'create_volume'},
     'account.tasks.enqueue_ready_volume': {'queue': 'enqueue_ready_volumes'},
 }
+CELERYBEAT_SCHEDULE = {
+    'scale_up_inspection_cluster_every_5_min': {
+        'task': 'account.tasks.scale_up_inspection_cluster',
+        'schedule': 60 * 5,  # seconds
```
ilyavw: Regarding timing, I think we originally were talking about trying to run the cluster once every 30 mins or an hour, maybe have it be configurable? Either way 5 minutes seems awfully small with how long snapshot copy operations can take.

cloudigrade/account/tasks.py
infinitewarp: info logs added in the latest commits. I figure start louder at info and we can lower it to debug when we decide it's too noisy.

infinitewarp:
Recorded demos here: https://github.com/cloudigrade/cloudigrade/issues/176#issuecomment-385073684

Issue #210, opened 2018-04-12
Celery waits too long if RabbitMQ cannot be reached


infinitewarp:
I encountered this issue locally when I forgot to bring up my RabbitMQ container. It seems that Celery waits an absurdly long time by default trying (unsuccessfully) to connect to the broker. This exceeded both nginx's and httpie's timeout. (nginx gives a 503 error)

I see this info in the docs for Redis:

http://docs.celeryproject.org/en/latest/getting-started/brokers/redis.html#visibility-timeout
> This option is set via the broker_transport_options setting:
>
> `app.conf.broker_transport_options = {'visibility_timeout': 3600}  # 1 hour.`
>
> The default visibility timeout for Redis is 1 hour.

But I didn't find a setting mentioned for RabbitMQ. Maybe the same setting applies?

If possible, we need a more reasonable value here, probably something on the order of single-digit seconds.

infinitewarp:
It looks like there are issues deeper in the Celery Kombu framework that cause this trouble. Unfortunately, it looks like (at least with the AMQP/RabbitMQ connection) we cannot fix this behavior without patching the internals of Kombu. What follows is a more detailed dump of my findings.

In the case of writing a messages to a queue using a `kombu.Producer`:
- `kombu.Producer.publish` takes `retry` bool and `retry_policy` dict as optional arguments.
- If `retry` is True, the `retry_policy` is passed into a decorator `kombu.Connection.ensure` that "Ensure operation completes. Regardless of any channel/connection errors occurring. Retries by establishing the connection, and reapplying the function."
- The `ensure` decorator takes the following optional arguments which are exploded from the aforementioned optional `retry_policy` dict given to the `publish` function:

            max_retries (int): Maximum number of times to retry.
                If this limit is exceeded the connection error
                will be re-raised.

            interval_start (float): The number of seconds we start
                sleeping for.
            interval_step (float): How many seconds added to the interval
                for each retry.
            interval_max (float): Maximum number of seconds to sleep between
                each retry.

- If `max_retries` is not specified, it retries forever.
- This all sounds great, but of course there's a catch! 🙁 
- Unfortunately,  at the following point deeper in the stack, `kombu.Connection.default_connection` calls `kombu.Connection.ensure_connection` **with no arguments** (even though `ensure_connection` has basically the same signature as `ensure`) which means that it will **always** iterate forever until it gets a valid connection.
![](https://i.imgur.com/WA7SXy3.png)

[...]

PR #239, opened 2018-04-30
Add health checks and update readiness/liveness probes.

ilyavw:


cloudigrade/config/urls.py
```diff
@@ -13,5 +13,6 @@
 urlpatterns = [
     url(r'^api/v1/', include(router.urls)),
     url(r'^api-auth/', include('rest_framework.urls')),
+    url(r'^healthz/', include('health_check.urls')),
```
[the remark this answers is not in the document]
ilyavw: https://stackoverflow.com/questions/43380939/where-does-the-convention-of-using-healthz-for-application-health-checks-come-f

cloudigrade/util/health.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class RabbitMQCheckBackend(BaseHealthCheckBackend):
+    """RabbitMQ Health Check."""
+
+    def check_status(self):
+        """Test RabbitMQ health check."""
+        conn = kombu.Connection(settings.RABBITMQ_URL)
+        try:
+            conn.connect()
```
infinitewarp: Given #210 (which _might_ be improved in a future release of Celery), this line might hang forever if RabbitMQ cannot be reached. As a newbie, I presume OpenShift will realize this is hanging and has its own timeout threshold that it will use to mark the service as critical/down, yes?

ilyavw: Yes, the readiness/liveness probes themselves have a timeout.

Issue #243, opened 2018-05-01
Celery worker missing from OpenShift environment


infinitewarp:
We missed this in #229 since the initial OpenShift environment PR merged very shortly before that one. We now have most things running in OpenShift, but we do not have a Celery worker running anywhere!

### Steps to reproduce

1. `make oc-up-all`
2. look at what's running in OpenShift

### Expected result

- There should be a container with a Celery worker to process tasks as they arrive.

### Actual result

- There are no containers running Celery workers!

ilyavw:
Looks good!

PR #248, opened 2018-05-01
add deployment config for cloudigrade-celery

infinitewarp:
https://github.com/cloudigrade/cloudigrade/issues/243

### DIFF
--- a/deployment/ocp/cloudigrade.yml
+++ b/deployment/ocp/cloudigrade.yml
@@ -10,7 +10,7 @@ objects:
   metadata:
     name: ${NAME}-celery
   spec:
-    replicas: 1
+    replicas: ${CELERY_REPLICA_COUNT}
     revisionHistoryLimit: 10
     selector:
       name: ${NAME}-celery
@@ -556,4 +556,9 @@ parameters:
 - name: RABBITMQ_PORT
   description: "Your rabbitmq port"
   value: "5672"
+  required: true
+# Celery worker settings
+- name: CELERY_REPLICA_COUNT
+  description: "Number of celery replicas"
+  value: "1"
   required: true
--- a/deployment/ocp/cloudigrade.yml
+++ b/deployment/ocp/cloudigrade.yml
@@ -8,7 +8,123 @@ objects:
 - kind: DeploymentConfig
   apiVersion: v1
   metadata:
-    name: cloudigrade
+    name: ${NAME}-celery
+  spec:
+    replicas: 1
+    revisionHistoryLimit: 10
+    selector:
+      name: ${NAME}-celery
+    template:
+      metadata:
+        labels:
+          name: ${NAME}-celery
+      spec:
+        containers:
+        - name: ${NAME}-celery
+          env:
+          - name: AWS_ACCESS_KEY_ID
+            valueFrom:
+              secretKeyRef:
+                key: aws-access-key-id
+                name: ${NAME}
+          - name: AWS_SECRET_ACCESS_KEY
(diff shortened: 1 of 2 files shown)


infinitewarp:
Demo... pull and run it yourself, or look over my shoulder? Here's a screenshot of how it appears to be running:

deployment/ocp/cloudigrade.yml
```diff
@@ -8,7 +8,123 @@ objects:
...
+    revisionHistoryLimit: 10
+    selector:
+      name: ${NAME}-celery
+    template:
+      metadata:
+        labels:
+          name: ${NAME}-celery
+      spec:
+        containers:
+        - name: ${NAME}-celery
```
ilyavw: This container is missing readiness and liveness probes, we need some method for checking if it's healthy.

deployment/ocp/cloudigrade.yml
```diff
@@ -8,7 +8,123 @@ objects:
 - kind: DeploymentConfig
   apiVersion: v1
   metadata:
-    name: cloudigrade
+    name: ${NAME}-celery
+  spec:
```
ilyavw: You do not have a a [deployment strategy](https://docs.openshift.com/dedicated/dev_guide/deployments/deployment_strategies.html) defined. Is celery compatible with a rolling deployment strategy (the default)? Or would using recreate here be better?

If you'd like to switch to using a recreate simply add 
```
strategy:
  type: Recreate
```
between the metadata and spec.

deployment/ocp/cloudigrade.yml
```diff
@@ -8,7 +8,123 @@ objects:
...
+    replicas: 1
```
ilyavw: Should this be configurable? Does this container support multiple pods running? Should we define [autoscaling](https://docs.openshift.com/dedicated/dev_guide/pod_autoscaling.html)?

deployment/ocp/cloudigrade.yml
infinitewarp: Does `strategy` belong *in* the `spec`, not as a peer? That's where it lives for the other `DeploymentConfig`s.

ilyavw: That's correct, my mistake!

infinitewarp:
@ilyavw I spent some time trying to get autoscaling to work per your review comment, but I time-boxed myself to end-of-today and didn't manage to get it working in that time. I hit a couple of problems along the way, although I'm willing to chock it up to ignorance on my part wrt configuration. 🙂 

- Functional autoscaling seems to require enabling metrics on the cluster, but I don't see how to do this. `oc cluster up` takes an optional `--metrics` flag (which is also marked "experimental"), but that didn't seem to enable what autoscaling needed.
- The necessary object `HorizontalPodAutoscaler` configuration requires `apiVersion: autoscaling/v1` and `apiVersion: extensions/v1beta1`. These seem to be fundamentally incompatible with the `apiVersion: v1` in our `cloudigrade.yml` `Template` definition, and I couldn't find a way to make them play nicely together. I found a workaround that suggested putting the `HorizontalPodAutoscaler` object into its own file to import, and that does seem to work but feels wrong.

Since I didn't get it totally working, I shoved my progress into another branch if we want to look at it or pick it up later: https://github.com/cloudigrade/cloudigrade/tree/243-worker-autoscaling

Also, since we don't have autoscaling on anything else yet but *eventually* probably want it on more things, I figure it's reasonable to punt on it for this PR and just get it working everywhere later.

PR #252, opened 2018-05-02
Detect when RabbitMQ is unreachable

### LINKED ISSUE #210: Celery waits too long if RabbitMQ cannot be reached

noahl:
Detect early if we can't reach RabbitMQ and raise an understandable
error message.

Closes #210 .

### DIFF
--- a/cloudigrade/account/exceptions.py
+++ b/cloudigrade/account/exceptions.py
@@ -3,3 +3,7 @@
 
 class InvalidCloudProviderError(RuntimeError):
     """Exception for when a report is run for an invalid cloud provider."""
+
+
+class RabbitMQUnreachableError(Exception):
+    """Exception for when we can't connect to RabbitMQ."""
--- a/cloudigrade/account/util.py
+++ b/cloudigrade/account/util.py
@@ -1,7 +1,7 @@
 """Various utility functions for the account app."""
 import collections
-from queue import Empty
 import socket
+from queue import Empty
 
 import kombu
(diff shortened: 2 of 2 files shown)


noahl:
@infinitewarp Thanks! I didn't think about the other ways kombu could be called.

Given our conversation in the sprint planning yesterday, I think we might want to drop this issue for now rather than keep looking for workarounds.

The next release of Kombu will be 4.2. They don't have a release timeline that I can see, but all of the issues in their 4.2 GitHub milestone are done except for actually making the release (https://github.com/celery/kombu/milestone/4), which I think is even more reason to not spend out time on a fix.

I'll close the PR and the issue and look for something else to do. If you think I should keep looking at this, we can always reopen.

PR #256, opened 2018-05-03
add header and section for new bugs to the default github issue template

infinitewarp:
I think we're overdue for updating our default issue template with a section for entering a bug.

It sure would be nice if GitHub supported picking from different templates under different circumstances, but I think this is a good enough compromise.

This is what the updated template looks like:
https://github.com/cloudigrade/cloudigrade/blob/even-better-git-issue-template/.github/ISSUE_TEMPLATE.md

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,9 +1,30 @@
+This is the default new issue template! You may use the next section to enter a new user story, or use the last section to enter a new bug. Please delete this header and the other unused section before submitting this issue.
+
+---
+
 As X, I want Y so that Z.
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,9 +1,30 @@
...
+1.
+1.
+
+## Expected Result
+
+-
+
+## Actual Result
+
+-
```
kdelee: I think one more section "Environment information". Could just be heading or could be a table with some default values:

service | provider | version
---------- | ----------- | --------------
cloudigrade | dev server | my-commit-hash
postgres | local openstack | default
rabbitmq | local openstack | default

PR #258, opened 2018-05-04
Change a dict to a set

### LINKED ISSUE #87: Be aware of windows instances/amis and act accordingly
As a Cloudigrade Administrator, I don't want cloudigrade wasting cycles trying to inspect windows instances/amis so that time could be used instead to inspect linux instances.

## Acceptance Criteria 

- [x] Verify that Windows instance AMIs are automatically stored in an inspected state and marked as a Windows image when they are discovered.

## Tech assumptions

- For AWS we can query the instance platform to find out if the instance is windows or not. Windows will be marked as `windows`, and everything else will be `None`.
```
$ aws ec2 describe-instances --query 'Reservations[*].Instances[*].[InstanceId,Platform,Status]' --output text
i-07ebccb011511a47c	None	None
i-0586c3c56cca7d93e	windows	None
i-0bee3c1c7c65b518c	None	None
```
- Further research will have to be done later to determine how to do this in other clouds.

noahl:
We had a dictionary that was only used for testing whether it
contained a certain key, never actually using the values. Make it
into a set instead.

### DIFF
--- a/cloudigrade/account/util.py
+++ b/cloudigrade/account/util.py
@@ -67,7 +67,7 @@ def create_new_machine_images(account, instances_data):
     """
     saved_amis = []
     for __, instances in instances_data.items():
-        platforms = {instance['ImageId']: instance['Platform']
+        platforms = {instance['ImageId']
                      for instance in instances
                      if instance.get('Platform', '').lower() == 'windows'}
         seen_amis = set([instance['ImageId'] for instance in instances])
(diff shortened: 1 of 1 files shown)


noahl:
Just rebased on top of #260 to fix the merge conflict.

noahl:
@infinitewarp I think I've addressed your comment, please take a look.

noahl:
Pushing a rebased version again because Brad merged #260.

Issue #259, opened 2018-05-04
get_running_instances only saves running instances in last reservation


infinitewarp:
While investigating https://github.com/cloudigrade/cloudigrade/issues/87 and doing some manual testing around Windows AMIs, I encountered an issue in which some running instances were not being saved upon account creation. Unfortunately, this is difficult to reproduce consistently because the issue is highly dependent on the order and grouping of instances on AWS's side, and we don't have a lot of control over that.

Specifically, when you call our `ec2.get_running_instances` for a region that has multiple instances, if those instances belong to *different* reservations, we only save the instances found in the *last* reservation. 

## Steps to Reproduce

1. Start multiple instances in a customer account on AWS.
1. Hope that AWS puts them in different reservations. 🙏 
1. If you have the customer's AWS profile set up locally, you can call `describe_instances` directly to see what instances are in what reservations: `AWS_PROFILE=customer aws ec2 describe-instances --query 'Reservations[*].Instances[*].InstanceId'`
1. Set up your environment for posting to cloudigrade (`make oc-up-all`, `ARN=`, `AUTH=`, …)
1. Register cloud account: `http post http://cloudigrade-myproject.127.0.0.1.nip.io/api/v1/account/ "${AUTH}" account_arn="${ARN}" resourcetype=AwsAccount`
1. Check the database for the expected instances: `oc rsh -c postgresql $(oc get pods -o jsonpath='{.items[*].metadata.name}' -l name=postgresql) scl enable rh-postgresql96 -- psql -c 'select * from account_awsinstance'`

## Expected Result

- Rows exist in the database for all running instances regardless of reservation.

## Actual Result

- Rows only exist for running instances in the *last* reservation.

infinitewarp:
demo of the fix: https://asciinema.org/a/179810

PR #260, opened 2018-05-04
fix to get instances from all reservations, not just the last one

infinitewarp:
fixes https://github.com/cloudigrade/cloudigrade/issues/259


infinitewarp:
demo: https://asciinema.org/a/179810

PR #264, opened 2018-05-07
instructions for aws customer to get arn by creating a policy and role

infinitewarp:
I just went through all the steps myself, and I wanted to get this documented somewhere that we could reference as we're defining the instructional needs of the UI.

FWIW, these docs **are not** intended to be considered final use for end users. This is effectively a first draft that I want to get posted so that we as developers and people defining the project have an understanding of what's needed of the customer in this process.

### DIFF
--- a/docs/aws-create-policy-and-role-for-arn.rst
+++ b/docs/aws-create-policy-and-role-for-arn.rst
@@ -24,17 +24,25 @@ Create Policy
                    "Sid": "VisualEditor0",
                    "Effect": "Allow",
                    "Action": [
-                       "ec2:DescribeImages",
                        "ec2:DescribeInstances",
+                       "ec2:DescribeImages",
+                       "ec2:DescribeSnapshots",
                        "ec2:ModifySnapshotAttribute",
-                       "ec2:DescribeSnapshotAttribute",
-                       "ec2:ModifyImageAttribute",
-                       "ec2:DescribeSnapshots"
+                       "ec2:DescribeSnapshotAttribute"
                    ],
                    "Resource": "*"
                }
            ]
        }
+
+   .. note::
+       cloudigrade needs each of these actions in order to identify and track relevant software products:
+
+       - `ec2:DescribeInstances <https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html>`_ enables cloudigrade to get information about your currently running instances.
+       - `ec2:DescribeImages <https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeImages.html>`_ enables cloudigrade to get information about the AMIs used to start your instances.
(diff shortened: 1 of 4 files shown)


docs/aws-create-policy-and-role-for-arn.rst
```diff
@@ -0,0 +1,63 @@
...
+                       "ec2:ModifyImageAttribute",
+                       "ec2:DescribeSnapshots"
+                   ],
+                   "Resource": "*"
+               }
+           ]
+       }
+
+#. Click "Review policy".
+#. Give the policy a distinct, memorable name such as ``policy-for-red-hat-cloud-meter``. Copy this name for reference because you will need it soon.
```
elyezer: I would avoid red hat and cloud meter here, that said I would suggest to name to `policy-for-cloudigrade` instead.

docs/aws-create-policy-and-role-for-arn.rst
```diff
@@ -0,0 +1,63 @@
...
+#. Click "Create policy".
+
```
elyezer: I think worth adding a note for each Action that snippet is allowing. Or point to AWS docs so user may review why those permissions are needed.

docs/aws-create-policy-and-role-for-arn.rst
```diff
@@ -0,0 +1,63 @@
...
+   #. log in to AWS console
+   #. search Services to go to IAM
+   #. click Roles in the left nav
+   #. click the "Create role" button
+   #. Click "Another AWS account"
+   #. Paste this number into the Account ID field: ``372779871274``
+
+#. Click "Next: Permissions".
+#. Enter your new policy's name in the search box (e.g. ``policy-for-red-hat-cloud-meter``)
```
elyezer: Again I would recommend naming to `policy-for-cloudigrade`

docs/aws-create-policy-and-role-for-arn.rst
```diff
@@ -0,0 +1,63 @@
...
+#. Check the box for the policy.
+#. Click "Next: Review".
+#. Give the role a distinct, memorable name such as ``role-for-red-hat-cloud-meter``.
```
elyezer: I would recommend naming to `role-for-cloudigrade`

docs/aws-create-policy-and-role-for-arn.rst
```diff
@@ -0,0 +1,63 @@
...
+#. Click "Create role".
+#. From the Roles list page, click the role name you just created.
+#. Copy the generated value for "Role ARN".
```
elyezer: I think a note explaining why this role is needed would be welcome. Better to be transparent, right?

infinitewarp:
Thanks for all the suggestions, @elyezer! I've taken the time here to clean up our required permissions and explain why we need each one.

To anyone else interested in this PR, please keep in mind that this is **not** intended to be final user-facing copy. This is effectively a first draft that I want to get posted so that we as developers and people defining the project have an understanding of what's needed of the customer in this process.

elyezer:
@infinitewarp thanks for the updates. That helps everyone trying to learn more about why those are needed, also it gives some clues about test cases we can explore, for example, what if a given permission is missing.

Issue #161, opened 2018-03-26
Restrict resource access by user


infinitewarp:
As an admin, I want Doppler to restrict access to resources and reports by their respective "owner" users so different users cannot see each others' data.

~Before we do this, we need to investigate how KeyCloak fits into the equation. See: https://github.com/cloudigrade/cloudigrade/issues/215~

**Update: Per Dan's decree, we should not be blocking on KeyCloak now. Better auth can come after the pilot launch.

## Acceptance Criteria 

- [x] Verify that creating a new cloud account (aka "clount") ties it to the active user (the authenticated user making the request).
- [x] Verify that listing accounts as user A includes *only* accounts created by user A and excludes accounts created by user B.
- [x] Verify that a superuser can list *all* accounts irrespective of who created them.
- [x] Verify that getting as user A an account created by user B results in a 404.
- [x] Verify that a superuser can get any individual account.
- [x] Verify that `make oc-user` continues to operate as-is to create a superuser.

## Assumptions and Questions

- Django has "is staff" versus "superuser"; we probably want "superuser" for now.
- To establish this relationship means we need to:
  - tie a Django user to the `Account` model
    - this means a 1-to-many relationship for User-to-Account
  - add queryset filtering in various places
    - with overrides for user.is_superuser
- For now, just filter at the account level and *not* denormalize the owner user across all the tables, but that could be an optimization later.
- To demo this, you probably should include the following steps:
  - create a new user (non-superuser)
  - create an account using that user's auth token
  - `http get` the account to see that the user (id may be good enough) is an attribute of the account

infinitewarp:
Demo the requested user_id: https://asciinema.org/a/180613

PR #272, opened 2018-05-09
restrict account resources by authenticated user

infinitewarp:
https://github.com/cloudigrade/cloudigrade/issues/161

This introduces some great changes to make authentication do something meaningful! _(finally 🎉 )_

1. it alters the model for `Account` to include a not-null foreign key to Django's `User`
2. it populates that new column with the currently-auth'd `User.id` at time of `Account` creation
3. it filters list and retrieve results from the `/account/` view to appropriate contents:
   - if regular user is auth'd, include only accounts belonging to that user
   - if superuser is auth'd, include all accounts regardless of user
   - if superuser is auth'd and does a list with query param `user_id=3` (for example), include accounts belonging specifically to the user with id `3`

### DIFF
--- a/cloudigrade/account/serializers.py
+++ b/cloudigrade/account/serializers.py
@@ -33,8 +33,12 @@ class AwsAccountSerializer(HyperlinkedModelSerializer):
             'id',
             'updated_at',
             'url',
+            'user_id',
         )
-        read_only_fields = ('aws_account_id',)
+        read_only_fields = (
+            'aws_account_id',
+            'user_id',
+        )
         extra_kwargs = {
             'url': {'view_name': 'account-detail', 'lookup_field': 'pk'},
         }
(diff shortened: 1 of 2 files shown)


cloudigrade/account/serializers.py
```diff
@@ -43,8 +43,12 @@ def create(self, validated_data):
         """Create an AwsAccount."""
         arn = aws.AwsArn(validated_data['account_arn'])
         aws_account_id = arn.account_id
-        account = AwsAccount(account_arn=str(arn),
-                             aws_account_id=aws_account_id)
+        user = self.context['request'].user
```
noahl: Can `user` be None here, if the request is unauthenticated?

infinitewarp:
New demo that includes @ilyavw 's requested user_id: https://asciinema.org/a/180613

Issue #273, opened 2018-05-09
create_volume task keeps retrying despite error state at AWS


infinitewarp:
When one of the operations we fire off to AWS as async produces an error, we should recognize that error, log and/or raise it somehow, and stop the backoff/retry mechanism. I recently experienced an issue where I was changing my AWS customer account's policy to remove permissions, and the async tasks running in the background kept running despite some errors. Specifically, I managed to time things such that cloudigrade successfully started the "copy snapshot" AWS command when I had removed access to it. This resulted in the new snapshot ID being created, but every time you described that new snapshot, AWS reported an error state.

How should we address this issue? In the various async tasks that check AWS objects for completion, we should specifically check for an "error" state and raise an appropriate exception that will stop the task from retrying.

## Steps to Reproduce

1. Not sure yet… will update this later.

## Expected Result

- Task runs once and halts when it encounters the error.

## Actual Result

- Task keeps retrying until it hits the configured stop limit.

ilyavw:
Verified.

PR #278, opened 2018-05-10
exit create_volume when snapshot has error

infinitewarp:
https://github.com/cloudigrade/cloudigrade/issues/273


infinitewarp:
Regarding the lack of reproduction steps in the linked issue, unfortunately I could not find a way to manually reproduce this issue. Originally this problem was revealed because I was experimenting with changing the behavior of cloudograde's calls to AWS (I had temporarily removed the `add_snapshot_ownership` call before the `create_volume` task), but since I abandoned those changes, actually reaching this condition seems to be more challenging.

If someone else can think of a way to reproduce this issue, I'd welcome the feedback. Otherwise, I'd ask reviewers to just give the unit tests a thorough reading and give this a pass/fail based on them.

PR #280, opened 2018-05-10
Add a backend for logins

noahl:
Use djoser to manage logins with a REST API. This commit just includes
a basic, functional backend. There is no user interface and no
customization of djoser.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -281,13 +281,18 @@ If your cloudigrade deployment failed because you didn't have ``AWS_ACCESS_KEY_I
 Authentication
 ==============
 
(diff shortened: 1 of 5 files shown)


elyezer:
> Also please include a small example of how someone can authenticate against the new endpoint. In the rest-api-examples.rst doc's "Authorization" section probably fits best that.

QE appreciates those examples :smile:

noahl:
@infinitewarp I just pushed a change that I think addresses your comment. Please take a look.

noahl:
@ilyavw the old commands should still work! I believe djoser is providing a REST interface to the same Django `User` objects that the standard Django tools use.

ilyavw:
@noahl I realize the old commands still work, but this seemed like a good time to start transitioning away from those as I see them eventually being phased out.

Issue #133, opened 2018-03-17
List/retrieve instances via REST


infinitewarp:
As an end user, I want to get my instances via REST so that I can see them in my web browser client.

## Acceptance Criteria 

- [x] Verify you can retrieve a single instance by id via http
- [x] Verify you can list all instances via http
- [x] Verify the list of instances is paginated
- [x] Verify user can only access instances that belong to accounts tied to that user ("who this user is" is based on the auth header)
- [x] Verify super user can view *all* instances.
- [x] Verify super user can filter instances to particular users. (e.g. query string param to indicate something like `user_id=foo`)
- [x] Verify you *cannot* create, edit, or delete an instance via http (post/put/patch/delete)
- [x] Verify endpoint is `/api/v1/instance/{id}`

## Assumptions and Questions

- Assumes we have implemented the user-account relationship (#161)
- This story doesn't have an explicit user-facing UI piece because we currently have no UI wireframes defined for viewing/listing events.

infinitewarp:
If you want to create some dummy instance data in your local cloudigrade, you can use `make oc-user` to create a superuser and use its credentials to get into [your local django admin](http://cloudigrade-myproject.127.0.0.1.nip.io/admin/) where you can manually create whatever objects you need.

kholdaway:
Demo --> https://asciinema.org/a/NsXQil5gg2apKsMIhb5Jwo5Y4

infinitewarp:
@kholdaway or @abaiken: would you kindly put your `demo.sh` in a gist and link it here?

kholdaway:
https://gist.github.com/kholdaway/bd27824e3a7c95c5abf2242a66f10987

PR #283, opened 2018-05-11
List/Retrieve instances via REST

abaiken:
Description: Enable users to access instances via REST. 
Issue: #133 
Demo: https://asciinema.org/a/NsXQil5gg2apKsMIhb5Jwo5Y4

### DIFF
--- a/cloudigrade/account/tests/test_views.py
+++ b/cloudigrade/account/tests/test_views.py
@@ -258,8 +258,7 @@ class InstanceViewSetTest(TestCase):
         )
         self.assertEqual(
             response.data['account'],
-            'http://testserver/api/v1/account/' +
-            str(instance.account.id) + '/'
+            f'http://testserver/api/v1/account/{instance.account.id}/'
         )
         self.assertEqual(
             response.data['resourcetype'], instance.__class__.__name__
--- a/cloudigrade/account/tests/test_views.py
+++ b/cloudigrade/account/tests/test_views.py
@@ -7,9 +7,9 @@ from rest_framework.test import APIRequestFactory, force_authenticate
 
 from account import AWS_PROVIDER_STRING
 from account.exceptions import InvalidCloudProviderError
-from account.models import AwsAccount
+from account.models import AwsAccount, AwsInstance
 from account.tests import helper as account_helper
-from account.views import AccountViewSet, ReportViewSet
+from account.views import AccountViewSet, InstanceViewSet, ReportViewSet
 from util.tests import helper as util_helper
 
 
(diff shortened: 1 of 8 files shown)


cloudigrade/account/views.py
```diff
@@ -1,10 +1,35 @@
...
-from account.models import Account, AwsAccount
+from account.models import (Account,
+                            AwsAccount,
+                            Instance)
+
+
+def convert_param_to_int(name, value):
+    """Check if a value is convertable to int.
+
+    :param name: The field name being validated
```
infinitewarp: Use Google style docstrings.

We may move to the RST style later, but consistency wins for now.

cloudigrade/account/views.py
```diff
@@ -25,7 +50,37 @@ def get_queryset(self):
...
+        if not user.is_superuser:
+            accounts = self.accounts.filter(user=user)
+            account_ids = [account.id for account in accounts]
+            return self.queryset.filter(account__in=account_ids)
+        user_id = self.request.query_params.get('user_id', None)
+        if user_id is not None:
+            user_id = convert_param_to_int('user_id', user_id)
+            accounts = self.accounts.filter(user__id=int(user_id))
+            account_ids = [account.id for account in accounts]
+            return self.queryset.filter(account__in=account_ids)
```
infinitewarp: I'm pretty sure there's a cleaner way of doing this that doesn't require the "in list" like this, but I'll have to play with this and come back…

infinitewarp: I played around in the shell with some local data, and this seems to work as desired:

```
>>> models.Instance.objects.filter(account__user__id=30)
<PolymorphicQuerySet [<AwsInstance i-0bee3c1c7c65b518c>, <AwsInstance i-05f78714782d7979b>]>
>>> models.Instance.objects.filter(account__user__id=31)
<PolymorphicQuerySet []>
```

So, I'd suggest simplifying `get_queryset` to look something like this:

```python
def get_queryset(self):
    """Get the queryset filtered to appropriate user."""
    user = self.request.user
    if not user.is_superuser:
        return self.queryset.filter(account__user__id=user.id)
    user_id = self.request.query_params.get('user_id', None)
    if user_id is not None:
        user_id = convert_param_to_int('user_id', user_id)
        return self.queryset.filter(account__user__id=user_id) 
    return self.queryset 
```

cloudigrade/account/views.py
kholdaway: Done.

cloudigrade/account/views.py
kholdaway: Great suggestion.  Done.

kholdaway:
@infinitewarp Have another look.  Thanks for the feedback.

cloudigrade/account/tests/test_views.py
```diff
@@ -221,3 +227,173 @@ def test_get_user1s_account_as_superuser_returns_ok(self):
...
+        self.factory = APIRequestFactory()
+
+    def assertResponseHasInstanceData(self, response, instance):
+        """Assert the response has data matching the instance object."""
+        self.assertEqual(
+            response.data['id'], instance.id
+        )
+        self.assertEqual(
+            response.data['account'],
+            'http://testserver/api/v1/account/' +
```
infinitewarp: This string concatenation could be collapsed to
```
f'http://testserver/api/v1/account/{instance.account.id}/'
```

abaiken:
@infinitewarp Done.

cloudigrade/account/views.py
```diff
@@ -1,10 +1,39 @@
...
 
 from account import serializers
 from account.exceptions import InvalidCloudProviderError
-from account.models import Account, AwsAccount
+from account.models import (Account,
+                            AwsAccount,
+                            Instance)
+
+
+def convert_param_to_int(name, value):
```
ilyavw: This isn't a view, perhaps move it to `account.util`?

ilyavw: If we plan to use it anywhere else, it's totally cool to move it to an appropriate place in `util.*`, maybe even a new package?

kholdaway:
@infinitewarp @ilyavw back at you

cloudigrade/account/views.py
```diff
@@ -1,10 +1,39 @@
...
+
+    """
+    if isinstance(value, int):
+        return value
+
+    try:
+        return int(value)
+    except ValueError:
+        error = {
+            name: [_('The id must be an integer.')]
```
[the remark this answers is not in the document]
kholdaway: That syntax didn't work but I did something similar.

cloudigrade/account/views.py
```diff
@@ -1,10 +1,39 @@
...
+        name (str): The field name being validated
+        value: The value to convert to int
+
+    Returns:
+        int: The int value
+    Raises:
+        ValidationError if value not convertable to an int
+
+    """
+    if isinstance(value, int):
```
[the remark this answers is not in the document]
kholdaway: 👍

kholdaway:
@infinitewarp Are the additional changes you need?

PR #284, opened 2018-05-12
Update issue templates

infinitewarp:


infinitewarp:
I experimented with [a personal repo](https://github.com/infinitewarp/github-goofs), and ZenHub's "create issue" button is unaware of the named/type templates, and it always uses the old default template. So, we'll want to keep the ugly old unified template around for anyone who wants to create issues that way.

infinitewarp:
@ilyavw FWIW, the commit I just pushed restored the old template file while keeping the new ones so the "new issue" view in ZenHub is not blank.

Issue #177, opened 2018-04-02
Run task to do inspection on cluster


infinitewarp:
As a Doppler admin, I want a Celery task to configure the cluster and launch "houndigrade" so that the actual inspection starts as soon as the cluster is ready.

Cluster configuration here includes:
- attach volumes to host (and build a list of ami-to-mountpath items)
- update the task definition
  - go to the container definition
  - update the command with appropriate CLI arguments
    - AMI/mount
    - the cloud type (AWS)

## Acceptance Criteria 

- [ ] Verify that when AWS says the cluster is not ready, this task aborts and retries later with exponential back-off.
- [ ] Verify that when AWS says the cluster is ready, this task does the appropriate cluster configuration and runs houndigrade.

## Tech assumptions

- Retry should use an exponential back-off [like this example](https://stackoverflow.com/a/9752811/6621995).
- This task should be called at the end when https://github.com/cloudigrade/cloudigrade/issues/176 succeeds.
- This task should take in its arguments a data structure representing *all* the volumes in the batch (for each: original AMI ID and volume ID).

ilyavw:
Currently blocked by `INC0728602`. We can not currently deploy a service object with type LoadBalancer which prevents us from verifying functionality and deployment.

PR #290, opened 2018-05-15
Add task to run inspection

ilyavw:
#177 

Ads a task that runs after the cluster is scaled up. Attached volumes to inspection machine, creates the task definition, and runs the task.

Demo: Soon™


ilyavw:
@infinitewarp the pypy3 job consistently fails, but I'm not quite sure why. If it _is_ based on python 3.5, it should have `assert_called_once()`.

infinitewarp:
FWIW, https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called

> New in version 3.6.

`assert_called` and `assert_called_once` are new in 3.6

`assert_not_called` is new in 3.5

That said… we talked offline and agreed we should just drop pypy3 support now.

cloudigrade/account/tasks.py
```diff
@@ -143,14 +146,84 @@ def scale_up_inspection_cluster():
...
+    result = ecs.describe_container_instances(
+        containerInstances=[result['containerInstanceArns'][0]],
+        cluster=settings.HOUNDIGRADE_ECS_CLUSTER_NAME
+    )
+    instance_id = result['containerInstances'][0]['ec2InstanceId']
+
+    # attach volumes
+    ec2 = boto3.resource('ec2')
+    for index, message in enumerate(messages):
+        mount_point = settings.HOUNDIGRADE_MOUNT_POINTS[index]
```
infinitewarp: Although we _should_ have `HOUNDIGRADE_AWS_VOLUME_BATCH_SIZE` equal to or less than the length of `HOUNDIGRADE_MOUNT_POINTS`, if either of those is misconfigured, this could result in an `IndexError`. How would you feel about checking that explicitly? Or maybe in the settings module itself do a quick sanity-check the those variables and blow up at startup if the batch size is ever greater than the number of mount points?

cloudigrade/config/settings/base.py
```diff
@@ -40,6 +40,64 @@
...
+    default='cloudigrade/houndigrade'
+)
+HOUNDIGRADE_ECS_IMAGE_TAG = env(
+    'HOUNDIGRADE_ECS_IMAGE_TAG',
+    default='latest'
+)
+HOUNDIGRADE_MOUNT_POINTS = env(
+    'HOUNDIGRADE_MOUNT_POINTS',
+    default=[
+        '/dev/xvdb',
```
infinitewarp: That looks like the right number of entries, which is a perfect match to the default `HOUNDIGRADE_AWS_VOLUME_BATCH_SIZE`, but I wonder if we should generate this list based on that value instead of 32 lines here. I don't feel strongly about this; just throwing out an idea.

ilyavw: List has now been replaced with `util.generate_device_name()`. Give it an index, and optionally a prefix, and it spits out a device name for you.

cloudigrade/account/tasks.py
ilyavw: Now that we no longer have a list, and generate the names on demand, I am resorting to trusting the `scale_up_inspection_cluster` to only give me enough messages up to the batch limit.

cloudigrade/config/settings/base.py
ilyavw: The spec for the device names on AWS, https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html#available-ec2-device-names . We're using HVM.

cloudigrade/account/tests/test_tasks.py
```diff
@@ -377,3 +378,66 @@ def test_scale_up_inspection_cluster_requeues_on_aws_error(
...
         )
         mock_run_inspection_cluster.delay.assert_not_called()
+
+    @patch('account.tasks.boto3')
+    def test_run_inspection_cluster_success(self, mock_boto3):
+        """Asserts successful starting of the houndigrade task."""
+        mock_list_container_instances = {
+            'containerInstanceArns': [util_helper.generate_dummy_instance_id()]
+        }
+        mock_ec2 = MagicMock()
```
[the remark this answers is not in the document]
ilyavw: Yup, MagicMock is needed for both to drastically simplify setup.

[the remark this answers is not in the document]
ilyavw: Actually I stand corrected, I only needed it for mock_ecs. Updating mock_ec2...

PR #292, opened 2018-05-15
Add Makefile commands to create users and superusers

noahl:


Makefile
```diff
@@ -130,14 +130,41 @@ oc-clean: oc-down
...
 
 user-authenticate:
 	@read -p "User name: " uname; \
 	$(PYTHON) $(PYDIR)/manage.py drf_create_token $$uname --settings=config.settings.local
 
 oc-user:
-	oc rsh -c cloudigrade $$(oc get pods -o jsonpath='{.items[*].metadata.name}' -l name=cloudigrade) scl enable rh-postgresql96 rh-python36 -- python manage.py createsuperuser
+	http post cloudigrade-myproject.127.0.0.1.nip.io/auth/users/create/ \
+	    username=user password=userpassword
+	oc rsh -c cloudigrade \
```
ilyavw: Is there a reason we're not using `https://github.com/cloudigrade/cloudigrade/blob/master/docs/rest-api-examples.rst#login-to-cloudigrade` here?

Makefile
```diff
@@ -130,14 +130,41 @@ oc-clean: oc-down
 	$(PREFIX) rm -rf $(OC_DATA_DIR)
 
 user:
-	$(PYTHON) $(PYDIR)/manage.py createsuperuser --settings=config.settings.local
+	http post localhost:8080/auth/users/create/ \
+	    username=user password=password
+	$(PYTHON) $(PYDIR)/manage.py drf_create_token user --settings=config.settings.local
```
ilyavw: Same as the `oc-user`, why not just call login as per `https://github.com/cloudigrade/cloudigrade/blob/master/docs/rest-api-examples.rst#login-to-cloudigrade`?

Makefile
```diff
@@ -130,14 +130,41 @@ oc-clean: oc-down
...
+	http post cloudigrade-myproject.127.0.0.1.nip.io/auth/users/create/ \
+	    username=user password=userpassword
+	oc rsh -c cloudigrade \
+	    $$(oc get pods -o jsonpath='{.items[*].metadata.name}' \
+	                   -l name=cloudigrade) \
+	    scl enable rh-postgresql96 rh-python36 -- \
+	    python manage.py drf_create_token user
+
+oc-superuser:
+	-oc rsh -c cloudigrade \
```
ilyavw: Why are we ignoring a failure here?

Makefile
```diff
@@ -130,14 +130,41 @@ oc-clean: oc-down
...
 user:
-	$(PYTHON) $(PYDIR)/manage.py createsuperuser --settings=config.settings.local
+	http post localhost:8080/auth/users/create/ \
+	    username=user password=password
+	$(PYTHON) $(PYDIR)/manage.py drf_create_token user --settings=config.settings.local
+
+superuser:
+	-$(PYTHON) $(PYDIR)/manage.py createsuperuser \
+	    --settings=config.settings.local \
+	    --username=admin --email=admin@admin.com --noinput
```
ilyavw: Let's use a "fake" domain like `example.com`, and not admin, we wouldn't want to accidentally send emails to this domain.

Makefile
noahl: I did it this way for consistency with the other commands, and also because `drf_create_token` has less noise in the output than the HTTP approach.

Makefile
noahl: Ignoring this failure lets you run the target twice and get a token each time, but only create a new user on the first run. I don't feel strongly about this either way.

Makefile
noahl: Good point. Done.

Makefile
```diff
@@ -130,14 +130,41 @@ oc-clean: oc-down
...
+	$(PYTHON) $(PYDIR)/manage.py drf_create_token admin --settings=config.settings.local
 
 user-authenticate:
 	@read -p "User name: " uname; \
 	$(PYTHON) $(PYDIR)/manage.py drf_create_token $$uname --settings=config.settings.local
 
 oc-user:
-	oc rsh -c cloudigrade $$(oc get pods -o jsonpath='{.items[*].metadata.name}' -l name=cloudigrade) scl enable rh-postgresql96 rh-python36 -- python manage.py createsuperuser
+	http post cloudigrade-myproject.127.0.0.1.nip.io/auth/users/create/ \
```
ilyavw: You can replace `cloudigrade-myproject.127.0.0.1.nip.io` here with `$$(oc get routes -o jsonpath='{.spec.host}' cloudigrade)/auth/users/create/` to always have the same route host as configured on the cluster.

Makefile
ilyavw: Maybe have a fetch token target instead? That way you can call it from the superuser targets and call it by itself when you just need a token.

Makefile
ilyavw: I'd prefer it if we had a clean split, if we're using the api to create a user, we should fetch the token the same way.

noahl: This concerns me because it's increasing the number of ways we do things, which means more ways things can go wrong.

How about standardizing on the API in all cases? That also means we dogfood our own APIs a bit more, which is always helpful.

ilyavw: The auth api is our standard. The token cli command is built into drf, and something that certainly will not be available in production, so we may as well get used to using the method that actual users would use.

Makefile
noahl: You're right, that's a better approach. Done, thanks!

Makefile
noahl: My mistake, I forgot that there was another reason. We can't use the standard login API because our users have no password. We have to use drf_create_token as a way around that.

Do you think it makes more sense to stick with drf_create_token, or create users with passwords?

noahl: I've been looking into creating users with passwords, but you really need Python for that. Given our discussion about keeping it simple, I'll just revert to drf_create_token.

noahl: To be clear, the regular users have passwords, but the superusers do not.

Making the retrieve token methods different would make life *really* difficult for anyone who wanted to build automation around these commands, though, so I strongly think they should work the same way.

noahl:
@ilyavw I pushed a new commit that I think addresses all of your comments.

infinitewarp:
Stepping back and trying to look at this holistically, I think we're overcomplicating the behaviors of the targets in duplicating functionality, baking in hard-coded values that limit the targets' reuse, and ultimately taking the makefile in a direction that doesn't _feel_ right.

Originally, we had `user` and `oc-user` that created a superuser locally and in openshift respectively. The contents of those targets didn't need changing and could have simply been renamed to `superuser` and `oc-superuser`. `user-authenticate` and `oc-user-authenticate` also need no changes; they already do exactly what they say. All we potentially needed here was a pair of commands `user` and `oc-user` to wrap up `http` requests to register a regular user, maybe taking inputs as `@read` to feed the call. They don't need followup commands to get tokens because we already have `user-authenticate` and `oc-user-authenticate` for that purpose.

That being said… we don't need to automate _everything_ in the makefile, and we shouldn't look at this like we're building out a CLI. As I think objectively about what really "belongs" in a makefile, I realize that these user management commands probably shouldn't be here at all, and it was probably a mistake to ask for them. Make targets should be simple helpers for developers to aid in the _setup and building_ of a project, not as a catch-all wrapper for any commands a developer might need. Since creating a regular user "properly" through the API is already so incredibly simple and not germane to getting cloudigrade running, I'm inclined to just drop this change and encourage developers to use the API directly as it's intended. We have docs for that, and the occasional one-liners are easy enough!

We probably should have been clearer at the start about the intent of the desired new targets, and I feel bad about maybe wasting your time on this one @noahl. Hopefully we've learned a thing or two from this exercise, and at least I think I have.

infinitewarp:
In that case, I'll close this one up…

PR #293, opened 2018-05-15
List/retrieve machine images via REST

### LINKED ISSUE #263: List/retrieve machine images via REST
As an end user, I want to get my machine images via REST so that I can see them in my web browser client.

## Acceptance Criteria 

- [x] Verify you can retrieve a single machine image by id via http
- [x] Verify you can list all machine images via http
- [x] Verify the list of machine images is paginated
- [x] Verify user can only access machine images that were used by instances that belong to accounts tied to that user ("who this user is" is based on the auth header)
- [x] Verify super user can view *all* images.
- [x] Verify super user can filter images to particular users. (e.g. query string param to indicate something like `user_id=foo`)
- [ ] Verify you *cannot* create or edit an image via http (post/put/patch)
- [x] Verify endpoint is `/api/v1/image/{id}`

## Assumptions and Questions

- Assumes we have implemented the user-account relationship (#161)
- This story doesn't have an explicit user-facing UI piece because we currently have no UI wireframes defined for viewing/listing events.

kholdaway:
Closes #263 
Demo -> https://asciinema.org/a/ZZjtL7xf2XzhEzXBYqqKNHUFM
Demo Source -> https://gist.github.com/kholdaway/5520eb74b292c37adcac7864ade61027

### DIFF
--- a/cloudigrade/account/tests/helper.py
+++ b/cloudigrade/account/tests/helper.py
@@ -125,7 +125,7 @@ def generate_aws_instance_events(
 
 
 def generate_aws_image(account,
-                       ec2_ami_id=None,
+                       ec2_ami_id,
                        is_encrypted=False,
                        is_windows=False):
     """
@@ -143,8 +143,7 @@ def generate_aws_image(account,
         AwsMachineImage: The created AwsMachineImage.
 
     """
-    if ec2_ami_id is None:
-        ec2_ami_id = helper.generate_dummy_image_id()
+    ec2_ami_id = helper.generate_dummy_image_id()
 
     return AwsMachineImage.objects.create(
         account=account,
--- a/cloudigrade/account/tests/helper.py
+++ b/cloudigrade/account/tests/helper.py
@@ -1,7 +1,10 @@
 """Helper functions for generating test data."""
 import random
 
-from account.models import (AwsAccount, AwsInstance, AwsInstanceEvent,
+from account.models import (AwsAccount,
+                            AwsInstance,
+                            AwsInstanceEvent,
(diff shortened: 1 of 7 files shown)


cloudigrade/account/tests/helper.py
```diff
@@ -119,3 +122,33 @@ def generate_aws_instance_events(
...
+                       ec2_ami_id=None,
+                       is_encrypted=False,
+                       is_windows=False):
+    """
+    Generate an AwsMachineImage for the AwsAccount for testing.
+
+    Any optional arguments not provided will be randomly generated.
+
+    Args:
+        account (AwsAccount): Account that owns the instance.
```
infinitewarp: `s/instance/image/`

cloudigrade/account/tests/helper.py
```diff
@@ -119,3 +122,33 @@ def generate_aws_instance_events(
...
+        ec2_ami_id (str): Optional EC2 ami id.
+        is_encrypted (bool): Optional Indicates if image is encrypted.
+        is_windows (bool): Optional Indicates if AMI is Windows.
+
+    Returns:
+        AwsMachineImage: The created AwsMachineImage.
+
+    """
+    if ec2_ami_id is None:
```
infinitewarp: codecov indicates that this conditional is _always_ executed and never skipped (and reading through the test code appears to confirms this). Since we never actually pass a value for `ec2_ami_id` info this function, we should just remove the parameter and drop the condition so the value is always generated by the helper.

kholdaway: 👍

Issue #212, opened 2018-04-12
List/retrieve events via REST


infinitewarp:
As an end user, I want to get my events via REST so that I can see them in my web browser client.

## Acceptance Criteria 

- [ ] Verify you can retrieve a single event by id via http
- [ ] Verify you can list all events via http
- [ ] Verify the list of events is paginated
- [ ] Verify user can only access events that belong to instances that belong to accounts tied to that user ("who this user is" is based on the auth header)
- [ ] Verify super user can view *all* events.
- [ ] Verify super user can filter events to particular users. (e.g. query string param to indicate something like `?user_id=foo`)
- [ ] Verify you *cannot* create or edit an event via http (post/put/patch)
- [ ] Verify endpoint is `/api/v1/event/{id}`
- [ ] Verify you can filter events by the instance that generated them (e.g. `?instance_id=123`)
- [ ] Verify that the `instance_id` and `user_id` filters stack appropriately

## Assumptions and Questions

- Assumes we have implemented the user-account relationship (#161)
- This story doesn't have an explicit user-facing UI piece because we currently have no UI wireframes defined for viewing/listing events.

abaiken:
Demo: [https://asciinema.org/a/msqRzOI5EythLSGwx0oCBo4Ys](https://asciinema.org/a/msqRzOI5EythLSGwx0oCBo4Ys)
Demo Source: [https://gist.github.com/abaiken/cd5560ab5a761c73a3d21a502a8a7637](https://gist.github.com/abaiken/cd5560ab5a761c73a3d21a502a8a7637)

infinitewarp:
Putting this back in review. Someone _other than_ the person authoring the code (ideally the person who entered the issue, but it doesn't have to be) needs to verify that the issue is resolved, check any acceptance criteria, and then close it.

infinitewarp:
I verified most of the AC, but the demo didn't cover these two:

- Verify you cannot create or edit an image via http (post/put/patch)
- Verify that the `instance_id` and `user_id` filters stack appropriately

Based on a review of the code, I believe the first one would be met since the new `InstanceEventViewSet` extends `viewsets.ReadOnlyModelViewSet`. I'm okay with this!

Also based on a review of the code, I believe the second one would be met, but we don't have a test case for it. Since this is the first of maybe more coming-soon API changes that will stack query string params as filters, I think we should add a test now so we can be confident that this stacking doesn't break with future changes.

[...]

PR #294, opened 2018-05-15
List/retrieve instance events via REST

abaiken:
Issue: #212 
Demo: [https://asciinema.org/a/msqRzOI5EythLSGwx0oCBo4Ys](https://asciinema.org/a/msqRzOI5EythLSGwx0oCBo4Ys)
Demo Source: [https://gist.github.com/abaiken/cd5560ab5a761c73a3d21a502a8a7637](https://gist.github.com/abaiken/cd5560ab5a761c73a3d21a502a8a7637)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,6 @@ jobs:
       install: pip install tox
       script: tox -eflake8
     - stage: test
-      python: pypy3
-      install: pip install tox
-      script: tox -epypy3
-    - stage: test
       if: (branch != master AND tag IS blank) OR (type IN (pull_request))
       services: docker
       sudo: required
--- a/README.rst
+++ b/README.rst
@@ -49,7 +49,7 @@ We encourage macOS developers to use `homebrew <https://brew.sh/>`_ to install a
 
     /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
     brew update
-    brew install python pypy3 gettext awscli postgresql socat
+    brew install python gettext awscli postgresql socat
     brew link gettext --force
     # We need to install a specific version of docker since newer ones have a bug around the builtin proxy
     brew cask install https://raw.githubusercontent.com/caskroom/homebrew-cask/61f1d33be340e27b91f2a5c88da0496fc24904d3/Casks/docker.rb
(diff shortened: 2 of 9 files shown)


cloudigrade/account/tests/helper.py
```diff
@@ -65,6 +65,49 @@ def generate_aws_instance(account, ec2_instance_id=None, region=None):
     )
 
 
+def generate_single_aws_instance_event(
+        instance, powered_time=None, event_type=None, ec2_ami_id=None,
+        instance_type=None, subnet=None):
+    """
+    Generate single AwsInstanceEvent for testing.
+
+    The ``powered_times`` defines when the instance should be considered
```
infinitewarp: This comment is incorrect. It references `powered_times` and says it's a tuple, but there is no `powered_times` in this function, and the nearest `powered_time` appears to be a `datetime`.

It looks like this was just blindly copied from `generate_aws_instance_events` but needs to be rewritten to be appropriate for this function.

abaiken: @infinitewarp Good catch! Fixed.

cloudigrade/account/tests/test_views.py
```diff
@@ -396,3 +399,195 @@ def test_list_instances_as_superuser_with_bad_filter(self):
...
+
+    def get_event_ids_from_list_response(self, response):
+        """
+        Get the instance event id values from the paginated response.
+
+        Args:
+            response (Response): Django response object to inspect
+
+        Returns:
+            set[int]: the instance id values found in the response
```
infinitewarp: `s/instance/event/`

abaiken: I need to pay more attention to my docstrings 🤦‍♀️ .

abaiken:
@infinitewarp I am not uncomfortable making that change! Should be easy enough!

abaiken:
@infinitewarp Updated with `instance_id` filter abilities. Also updated the demo & gist.

cloudigrade/account/views.py
```diff
@@ -73,11 +73,17 @@ def get_queryset(self):
         """Get the queryset filtered to appropriate user."""
         user = self.request.user
         if not user.is_superuser:
-            return self.queryset.filter(instance__account__user__id=user.id)
+            self.queryset = self.queryset.filter(
```
infinitewarp: This looks like it's generally on the right track, but we should assign the filtered queryset to a local variable (perhaps just `queryset`) instead of overriding `self.queryset` and return that local variable at the end. By clobbering the instance variable as is currently here, later calls to `get_queryset` will already have these filters already applied at the start which may be wrong depending on their context.

abaiken: @infinitewarp Fixed.

Issue #291, opened 2018-05-15
Celery worker doesn't process scheduled tasks if also configured to listen to a queue.


ilyavw:
## Summary

If I have my celery worker configured in such a way that it listens to queues and runs scheduled tasks, it'll fail to run scheduled tasks.

## Steps to Reproduce

1. Run your celery worker with `-l info -A config worker --beat --scheduler django_celery_beat.schedulers:DatabaseScheduler -Q copy_ami_snapshot,create_volume,enqueue_ready_volumes`
2. Register a new account and let it start copying volumes.

## Expected Result

- `scale_up_inspection_cluster()` runs

## Actual Result

- `scale_up_inspection_cluster()` never runs

## Additional context

If you run celery without the -Q params, ie `-l info -A config worker --beat --scheduler django_celery_beat.schedulers:DatabaseScheduler` the `scale_up_inspection_cluster()` task runs without issue.

infinitewarp:
Possibly related anecdote:

When @adberglund and I first started using Celery locally with cloudigrade, we found that Celery's documentation indicates that _by default_ all known queues should be enabled for the `worker` command, and `-Q` should only be necessary if you want to enable specific queues (see [documentation](http://docs.celeryproject.org/en/latest/reference/celery.bin.worker.html#cmdoption-celery-worker-q). However, we could never actually get this to work. Instead, as you see now in our deployment configs, we've had to manually specify _all_ of the task queues we want the worker to care about.

This has bad code smell all over it, but we haven't taken the time to investigate it yet.

There's a good chance we misunderstood something or misconfigured the way we have our tasks mapped to queues. So, whoever picks this bug up should keep that in mind and see if there's anything related to our task configuration that could be affecting this bad behavior with the beat scheduler.

noahl:
The bug seems to be that the autoscaling task is routed to the default queue `celery` (because no other queue is explicitly configured), but the `celery` queue is not enabled for the worker, so the messages are never received.

I tested it by standing up a cloudigrade cluster with the current master branch code. The logs included `[2018-05-17 19:12:18,082: INFO/Beat] Scheduler: Sending due task scale_up_inspection_cluster_every_60_min (account.tasks.scale_up_inspection_cluster)`, but nothing else about scaling up.

I also ran `rabbitmqctl list_queues name messages_ready messages_unacknowledged` in the celery container, and got
```rabbitmqctl list_queues name messages_ready messages_unacknowledged
Timeout: 60.0 seconds ...
[...]

PR #298, opened 2018-05-17
Route autoscale tasks to their own queue

noahl:
https://github.com/cloudigrade/cloudigrade/issues/291

And add the queue to the celery worker configuration

### DIFF
--- a/cloudigrade/config/settings/base.py
+++ b/cloudigrade/config/settings/base.py
@@ -223,6 +223,8 @@ CELERY_TASK_ROUTES = {
     'account.tasks.copy_ami_snapshot': {'queue': 'copy_ami_snapshot'},
     'account.tasks.create_volume': {'queue': 'create_volume'},
     'account.tasks.enqueue_ready_volume': {'queue': 'enqueue_ready_volumes'},
+    'account.tasks.scale_up_inspection_cluster':
+        {'queue': 'scale_up_inspection_cluster'},
 }
 CELERY_BEAT_SCHEDULE = {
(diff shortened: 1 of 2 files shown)


noahl:
I tested this PR by manually triggering an autoscale like this:
```oc rsh cloudigrade-celery-1-cfrgf scl enable rh-postgresql96 rh-python36 -- bash
$ ./manage.py shell
>>> import account.tasks
>>> account.tasks.scale_up_inspection_cluster.delay()
```

After this, the logs for the cloudigrade container contained:
```[2018-05-17 20:45:16,605: INFO/ForkPoolWorker-3] Starting new HTTPS connection (1): autoscaling.us-east-1.amazonaws.com
[2018-05-17 20:45:16,861: INFO/ForkPoolWorker-3] Not scaling up because no new volumes were found.
[2018-05-17 20:45:16,861: INFO/ForkPoolWorker-3] Task account.tasks.scale_up_inspection_cluster[e945ea61-c8b8-457d-a80c-ba3a58568fb8] succeeded in 0.2967632889995002s: None
```

I tried to catch the task in RabbitMQ using `rabbitmqctl`, but it was completed so quickly I never saw it.

ilyavw:
Overall this looks fine to me, one thing I'm curious about though is if we can move all the celery command parameters into a config file? I believe that should give us one place to configure the worker and hopefully keep this issue from coming up again.

noahl:
@ilyavw I think we could have either a separate file with celery settings or a `SimpleNamespace` object to group them all together. I really like the idea, but I think it's a separate change than this PR. I'd be happy to do it next.

Issue #286, opened 2018-05-14
500 server error when attempt to create account (API)


kdelee:
## Summary

When I attempt to create an account with a `POST` to `api/v1/account/`, I get a 500 internal server error instead of some more informative error that tells me what I did wrong.

## Steps to Reproduce

Send a `POST` to `/api/v1/account/` using proper authentication with data:

```
'{"resourcetype": "AwsAccount", "account_arn": "arn:aws:iam::518028203513:role/grant_cloudi_to_372779871274"}'
```

(I got this arn from the api example docs)

## Expected Result

- Either 200 success and a new account or a 400 with an error message that tells me what I did wrong.

## Actual Result

```
response code : 500

text_error_message : '<h1>Server Error (500)</h1>'
```

## Additional context

Serving locally in openshift, using 5428f7cbffde008698bb0de8153f1123d237ec2c .
Was attempting to create accounts with API in order to test ability to list them.

This is blocking me from testing listing the accounts because I would like to do the following:

1) Create X number of accounts
2) Request list of accountds
3) Validate all the accounts are in the list

kdelee:
@infinitewarp another thing is that a `POST` to `api/v1/account/` is only told one item that is missing from the request. In my experience it is possible and preferable to return a list of all missing items. So right now if I send an empty body with a `POST` I only get back the message: `{'resourcetype': 'This field is required'}` but I should get back `[{'resourcetype': 'This field is required'}, {'account_arn': ['This field is required.']} ]` or something like that.

Do you want to roll that into this as it is a part of the `400` response I should (perhaps) be getting if I'm doing something wrong, or do you want a separate issue?

Side note: if endpoints could report the fields they accept when you send `OPTIONS` to the endpoint that would be a great way to make the API more discoverable for QE and UI devs. i.e. right now I'm guessing that `resourcetype` has a finite list of valid values (`['AwsAccount']`)

kdelee:
@infinitewarp 

Recreated again this morning with fresh environment. My set up was this:

1) Pull lastest master branch
2) make oc-up-all
3) make oc-user
4) make oc-user-authenticate & save token
5) `http post http://$CLOUDIGRADE_BASE_URL/api/v1/account/ Authorization:"Token ${CLOUDIGRADE_TOKEN}" resourcetype="AwsAccount" account_arn="arn:aws:iam::518028203513:role/grant_cloudi_to_372779871274"`

infinitewarp:
[...]

PR #300, opened 2018-05-18
286 better handle bogus arns

infinitewarp:
https://github.com/cloudigrade/cloudigrade/issues/286

This change actually does a couple of things:

- Explicitly handle if we don't have access to the provided ARN for account creation.
- If we have access to the ARN but the policy isn't good, expand the response with more meaningfully verbose explanation of the failures.
- In _general_ handle any non-DRF exceptions and present the simplified DRF "500 error" response so we don't leak a bunch of Django internals.

Demo: https://asciinema.org/a/182337

### DIFF
--- a/cloudigrade/account/serializers.py
+++ b/cloudigrade/account/serializers.py
@@ -1,6 +1,7 @@
 """DRF API serializers for the account app."""
 import logging
 
+from botocore.exceptions import ClientError
 from dateutil import tz
 from django.db import transaction
 from django.utils.translation import gettext as _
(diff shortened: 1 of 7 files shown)


infinitewarp:
Demo link: https://asciinema.org/a/182337

infinitewarp:
Since It looks like I've got the initial approval from you guys, I'm going to go ahead and merge this PR since this should at least improve the situation. If the originating bug is still reproducible in some ways that I haven't found, we can address those in future commits and PRs.

Issue #237, opened 2018-04-30
Authenticate via web browser


infinitewarp:
As an end user, I want to be able to log in via my browser so that I can access my data.

## Acceptance Criteria 

- [x] Verify user can log in using web browser.
- [x] Verify documentation is updated if necessary to explain how auth token/header works.
- [x] Verify documentation exists to explain how to authenticate.
- [x] Verify integration tests's issue exists (see https://github.com/cloudigrade/integrade/issues/22)

## Assumptions and Questions

- This assumes that SOMEHOW a regular, non-superuser user exists in the DB.
  - This leaves it to the developer of this story to inject that data and explain.
  - This could be just a SQL `INSERT` statement, use Django built-in admin, etc.
- This requires a basic working web client with a login form.
- This would be an excellent candidate for a first _real_ integration test.
- For the backend, please look at:
  - https://github.com/sunscrapers/djoser
  - https://github.com/Tivix/django-rest-auth
  - http://www.django-rest-framework.org/api-guide/authentication/
- If using a third-party framework gets additional functionality like user registration or password reset working as well, great! 👍

noahl:
https://github.com/cloudigrade/cloudigrade/pull/280 has the backend code for this

noahl:
Here's a demo for the backend-only PR: https://asciinema.org/a/SVspdvpVGWrzXnOp4HAY8k1uC

Since this is partial work and the verification criteria above don't really apply, here's what I did:

1. Create user

http post http://cloudigrade-myproject.127.0.0.1.nip.io/auth/users/create/ username=developer password=djoserpassword

2. Try to get user info

http get http://cloudigrade-myproject.127.0.0.1.nip.io/auth/me/

3. Try to list accounts (or any other protected API endpoint)

http get http://cloudigrade-myproject.127.0.0.1.nip.io/api/v1/account/

4. Login as user

http post http://cloudigrade-myproject.127.0.0.1.nip.io/auth/token/create/ username=developer password=djoserpassword

export AUTH='Authorization: Token ...'

5. Get user info

http get http://cloudigrade-myproject.127.0.0.1.nip.io/auth/me/ "${AUTH}"

6. List accounts

http get http://cloudigrade-myproject.127.0.0.1.nip.io/api/v1/account/ "${AUTH}"

7. Logout as user

http post http://cloudigrade-myproject.127.0.0.1.nip.io/auth/token/destroy/ "${AUTH}"

8. Try to get user info with old token

[...]

PR #308, opened 2018-05-24
Update nginx.conf with localhost cors magic.

ilyavw:
Updates #237 

This adds CORS headers when your origin is either on `.127.0.0.1.nip.io` or `localhost:*`.

### DIFF
--- a/deployment/nginx/nginx.conf
+++ b/deployment/nginx/nginx.conf
@@ -12,6 +12,15 @@ http {
     default_type application/octet-stream;
     access_log /var/opt/rh/rh-nginx112/log/nginx/access.log combined;
 
+    map $http_origin $DO_CORS {
+        hostnames;
+
+        default 'false';
+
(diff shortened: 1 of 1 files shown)


deployment/nginx/nginx.conf
```diff
@@ -12,6 +12,15 @@ http {
     default_type application/octet-stream;
     access_log /var/opt/rh/rh-nginx112/log/nginx/access.log combined;
 
+    map $http_origin $DO_CORS {
+        hostnames;
```
infinitewarp: It's been a while since I had my head in the nginx config space. What does `hostnames;` do here?

ilyavw:
```
hostnames
indicates that source values can be hostnames with a prefix or suffix mask
```
http://nginx.org/en/docs/http/ngx_http_map_module.html

ilyavw:
https://asciinema.org/a/mDb3hycLEifopVQLZglwxhNw5

Issue #296, opened 2018-05-16
Spike: Experiment swapping RabbitMQ with SQS


infinitewarp:
Unfortunately, where we currently plan to run cloudigrade for test and production in [OpenShift Dedicated](https://console.insights.openshift.com/console/project/cloudigrade-test/overview) appears to be configured in such a way that we cannot expose RabbitMQ's TCP port outside of the local cluster ([details](https://github.com/cloudigrade/cloudigrade/issues/177#issuecomment-388058966)). This means the houndigrade cluster in AWS cannot currently reach our shared RabbitMQ instance as currently designed.

One of the proposed solutions to this problem is to adapt cloudigrade to use Amazon SQS where it currently uses RabbitMQ. Since we use Celery, and [Celery appears to support Amazon SQS as a broker](http://docs.celeryproject.org/en/latest/getting-started/brokers/index.html), this _could be_ as simple as a configuration change! :tada:

The goals of this spike are to:
- experiment locally with configuring cloudigrade and the Celery workers to use SQS queues
- verify that the task and message queue functionalities we need work as expected
- document your findings here including any necessary changes, limitations, and bonus features
- if successful, submit a PR with appropriate configuration changes so that test and future prod environments can switch to using Amazon SQS queues

Bonus points if you have time:
- Look into [localstack](https://localstack.cloud/) as a way to have AWS-like infrastructure locally

If this is as easy as making configuration changes, then individual developers can continue using local RabbitMQ instances _for now_ even though we are aware that this prevents developers from receiving results back from houndigrade. We may resolve that problem separately when we provision more AWS accounts for everyone on the team to use.

Time box: 2 days

infinitewarp:
I'm pulling this back into the "In Progress" state while I work on some issues around credentials for houndigrade and write some documentation for developers to understand what environment changes are necessary.

infinitewarp:
Demo: https://asciinema.org/a/185696

infinitewarp:
To answer the question about localstack in the original description, no, that project is too immature and doesn't support all the things we would need.

PR #312, opened 2018-05-25
Replace RabbitMQ with Amazon SQS

infinitewarp:
#296 

This PR changes cloudigrade to use Amazon SQS exclusively for its Celery tasks and queues. Another parallel change needs to be made to houndigrade to complete the replacement.


infinitewarp:
Good catch @elyezer, and that reminds me I should update the documentation to explain how to configure and get to SQS. I'll submit another updated commit with those changes soon.

infinitewarp:
See also: https://github.com/cloudigrade/houndigrade/pull/12

README.rst
```diff
@@ -314,7 +320,7 @@ This auth token can be supplied in the Authorization header.
 Message Broker
 ==============
 
-RabbitMQ is used to broker messages between **cloudigrade** and inspectigrade services. There are multiple Python packages available to interact with RabbitMQ; the officially recommended packaged is `Pika <https://pika.readthedocs.io/en/latest/>`_. Both services serve as producers and consumers of the message queue.
+Amazon SQS is used to broker messages between **cloudigrade** and inspectigrade services.
```
kdelee: `s/inspectigrade/houdigrade/` ???

README.rst
```diff
@@ -139,11 +139,17 @@ Use the AWS CLI to save that configuration to your local system:
...
+AWS access for running **cloudigrade** inside a local OpenShift cluster must be enabled via environment variables. Set the following variables in your local environment *before* you start running in OpenShift. Values for these variables can be found in the files in your ``~/.aws/`` directory.
 
 -  ``AWS_ACCESS_KEY_ID``
 -  ``AWS_SECRET_ACCESS_KEY``
 -  ``AWS_DEFAULT_REGION``
+-  ``AWS_SQS_ACCESS_KEY_ID``
+-  ``AWS_SQS_SECRET_ACCESS_KEY``
+-  ``AWS_SQS_REGION``
+-  ``AWS_SQS_QUEUE_NAME_PREFIX``
+
```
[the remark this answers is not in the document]
infinitewarp: Yes, I believe that's all we need. The `AWS_SQS_QUEUE_NAME_PREFIX` makes it so that each of us have our own pseudo-name-spaced queues (we each have _several_) since we're all sharing one master AWS account. If someone doesn't set an appropriate value here, that person risks colliding with another developer, and this is where I've asked people to "play nicely" with this limitation. :)

kholdaway:
Looks good to me.  After I watch the demo.

infinitewarp:
Demo: https://asciinema.org/a/185696

tox.ini
```diff
@@ -25,6 +25,11 @@ deps =
 setenv =
   DJANGO_SETTINGS_MODULE=config.settings.test
   PYTHONPATH={toxinidir}/cloudigrade
+passenv =
+  AWS_SECRET_ACCESS_ID
```
ilyavw: How come we need real values for tox?

infinitewarp: I have this because the houndigrade script still needs to be able to talk to SQS to send its results, and passing these through individually is easier than trying to construct and pass AWS_SQS_URL since you'd have to url-escape the key and secret in order to build it. I actually started down that road before, and I gave up when the escapes got ugly.

I'm happy to entertain alternate solutions if you have one. 🙂

infinitewarp: Good point. I've replaced these with bogus values in the latest commit. We need some values here because of how we construct the SQS URL in the settings, and we can't just set defaults in `test.py` because they need to be present when `base.py` loads which happens _before_ `test.py`.

tox.ini
```diff
@@ -25,11 +25,10 @@ deps =
 setenv =
   DJANGO_SETTINGS_MODULE=config.settings.test
   PYTHONPATH={toxinidir}/cloudigrade
-passenv =
-  AWS_SECRET_ACCESS_ID
-  AWS_SECRET_ACCESS_KEY
-  AWS_SQS_ACCESS_KEY_ID
-  AWS_SQS_SECRET_ACCESS_KEY
+  AWS_SECRET_ACCESS_ID=foo
```
ilyavw: Can we just put these in `config.settings.test` so tox doesn't accidentally slurp in credentials from the env? (I assume it'll use either bogus creds if one isn't present in the env)

infinitewarp: See my previous comment. 

> we can't just set defaults in `test.py` because they need to be present when `base.py` loads which happens before `test.py`.

ilyavw: Sounds like the problem we had before with rabbitmq url, in which case we just copied those bits to `test.py`. While that does "duplicate" code, it's a known problem, vs potentially adding a new one to the tox file.

Or am I misunderstanding something?

infinitewarp: This is different because now in base we need credentials to be present when we get the frozen credentials from the boto session so we can ensure we can build the SQS URL. 

I could move that functionality out into a new “dev/local”.py settings file and then it could be duplicated and changed in test and prod settings as needed, but that seems a little bit overkill to avoid a few tox environment variables since the variables have to exist in every other non-test environment anyway.

Issue #317, opened 2018-05-29
Respond to unauthenticated OPTIONS requests


noahl:
This is required by the W3C (https://www.w3.org/TR/cors/#cross-origin-request-with-preflight-0), but Django has an open issue (https://github.com/encode/django-rest-framework/issues/5616) which will require a workaround.

---

As a developer, I want Cloudigrade to respond to unauthenticated OPTIONS requests so I can use CORS with preflight requests.

## Acceptance Criteria

- [x] Verify that `/auth/me/` (specific URL chosen for concreteness) responds to an unauthenticated OPTIONS request with a `2XX` response code.
- [x] Verify that the permissions change will propagate to all Cloudigrade URLs (can be done by using the `DEFAULT_PERMISSION_CLASSES` workaround in the Django issue above).

## Steps to Reproduce

1. Make an unauthenticated OPTIONS request to `/auth/me/` on a Cloudigrade server.

## Expected Result

- Get an HTTP response with response code in the 200s.

## Actual Result

- Get a 401 permission error.

noahl:
As I'm about to head out the door, my work so far is in the branch `317-unauthenticated-options` on GitHub. I've added a test to check that permissions work (which fails on the current code) and I have a new permissions class that I think will make things work (from a workaround at https://github.com/encode/django-rest-framework/issues/5616).

However, I've discovered that Djoser hard-codes the default DRF permissions classes and doesn't offer a hook to change, which has lead me down the road of trying to monkey-patch djoser. I spent most of the day fighting a circular import issue, but now it looks like there's some sort of bug in the authentication code itself. The latest test results are at https://travis-ci.org/cloudigrade/cloudigrade/jobs/386331381 .

infinitewarp:
@ilyavw 's demo for his workaround fix in #324: https://asciinema.org/a/oiTx4MQlluaLts36IJSe5byoH

It looks like that should resolve this issue.

PR #324, opened 2018-06-04
Add workaround for drf/djoser handling of options calls

ilyavw:
Updates #317 

This adds a workaround for djoser explicitly defining which permissions classes it uses for specific views.

I've excluded certain parts of the code from testing/linting as it was lifted directly from djoser, parts that were added (`apps.py, permissions.py`) are all tested.


ilyavw:
Quick demo: https://asciinema.org/a/oiTx4MQlluaLts36IJSe5byoH

PR #326, opened 2018-06-05
Persist houndigrade results as they are pulled off the queue

kholdaway:


kholdaway:
Demo video -> https://drive.google.com/open?id=1NnNh71gnDy0G6d0Z4uzrVuOS_WgXU55B

cloudigrade/account/migrations/0007_add_image_properties.py
```diff
@@ -0,0 +1,73 @@
...
+    """Copy old is_windows value to an image tag."""
+    ContentType = apps.get_model('contenttypes', 'ContentType')
+
+    # Create image tags
+    ImageTag = apps.get_model('account', 'ImageTag')
+    image_tag_ct = ContentType.objects.get_for_model(ImageTag)
+
+    windows_tag = ImageTag.objects.create(
+        description='windows', polymorphic_ctype=image_tag_ct)
+    ImageTag.objects.create(description='rhel', polymorphic_ctype=image_tag_ct)
```
infinitewarp: I don't think we need the polymorphic stuff on the `ImageTag` class; I'm struggling to think of when it might be sub-classed since the data we're putting there should be cloud-agnostic.

You're getting the `PolymorphicModel` inheritance chain here that requires that extra kwarg because `ImageTag` is extending `BaseModel`. I think it would be a good idea now to create perhaps _two different_ "base" model classes: one that is and one that isn't polymorphic. In `models.py` they would basically have the exact same definition except for their name and the classes they extend.

Here's a breakdown of what I'm thinking we'd change, and I think it should be an easy clean shuffle:

- Rename `BaseModel` to `BasePolymorphicModel`.
- Change `Account`, `Instance`, `InstanceEvent`, and `MachineImage` to extend `BasePolymorphicModel`.
- Create a new `BaseModel` that simply extends `django.db.models.Model` and replicates the `Meta`, `created_at`, and `updated_at` bits from `BasePolymorphicModel`. (Could share a mixin? I'm not sure about that, though, because using mixins on Django models to change `Meta` can do unexpected things. It might just have to be a little bit of duplicated code.)
- Change `ImageTag` to extend `BaseModel`.
- Remove the unnecessary `polymorphic_ctype`s on the `ImageTag` object creations.

infinitewarp: (This is of course assuming that being polymorphic doesn't _become_ necessary as some side-effect of interacting with the other polymorphic `MachineImage` class. I confess I haven't actually tried this myself.)

kholdaway: I will give it a try and let everyone know!

cloudigrade/account/tasks.py
```diff
@@ -228,3 +235,54 @@ def run_inspection_cluster(messages, cloud='aws'):
...
+            rhel_found = any([attribute.get('rhel_found')
+                              for disk_json in list(image_json.values())
+                              for attribute in disk_json.values()])
+            if rhel_found:
+                ami.tags.add(RHEL_TAG)
+            # Add image inspection JSON
+            ami.inspection_json = json.dumps(image_json)
+            ami.save()
+        else:
+            logger.error(_('AwsMachineImage %s is not found.' % image_id))
```
ilyavw: We've generally been using .format for our strings, in addition this won't actually work with gettext since it'll try to look it up with the image_id in the string. Try changing it to:
```
logger.error(_('AwsMachineImage "{0}" is not found.').format(image_id))
```

cloudigrade/account/tasks.py
```diff
@@ -228,3 +235,54 @@ def run_inspection_cluster(messages, cloud='aws'):
...
+        HOUNDIGRADE_MESSAGE_READ_LEN)
+    if bool(messages):
+        for message in messages:
+            inspection_result = message
+            if isinstance(message, str):
+                inspection_result = json.loads(message)
+            if inspection_result.get(CLOUD_KEY) == CLOUD_TYPE_AWS:
+                persist_aws_inspection_cluster_results(inspection_result)
+            else:
+                logger.error(_('Unsupported cloud type: %s' %
```
ilyavw: Same as line 264 above.

tox.ini
```diff
@@ -30,7 +30,7 @@ setenv =
   AWS_SQS_ACCESS_KEY_ID=fizz
   AWS_SQS_SECRET_ACCESS_KEY=buzz
 commands =
-  coverage run {toxinidir}/cloudigrade/manage.py test account analyzer dj_auth util
+  coverage run {toxinidir}/cloudigrade/manage.py test {posargs: account analyzer dj_auth util}
```
ilyavw: What does adding `{posargs}` here get us? I've never used it so would love to know!

kholdaway: This allows you pass in a test case you want to run.  For example `tox -e py36 -- account.tests.test_tasks.AccountCeleryTaskTest.test_persist_inspect_results_aws_cloud_image_not_found` will only run that test.  Must quicker debug cycle when developing new tests.

infinitewarp: That's a neat feature! I've never seen it done like that before either.

cloudigrade/account/tasks.py
```diff
@@ -228,3 +235,55 @@ def run_inspection_cluster(messages, cloud='aws'):
...
+    if bool(messages):
+        for message in messages:
+            inspection_result = message
+            if isinstance(message, str):
+                inspection_result = json.loads(message)
+            if inspection_result.get(CLOUD_KEY) == CLOUD_TYPE_AWS:
+                persist_aws_inspection_cluster_results(inspection_result)
+            else:
+                logger.error(_('Unsupported cloud type: "{0}"'.format(
+                               message.get(CLOUD_KEY))))
```
infinitewarp: Parentheses aren't nested correctly here for formatting the localized string.

Should be:
```
logger.error(_('Unsupported cloud type: "{0}"').format(message.get(CLOUD_KEY))) 
```

cloudigrade/account/migrations/0007_add_image_properties.py
```diff
@@ -0,0 +1,62 @@
...
+    """Copy old is_windows value to an image tag."""
+
+    # Create image tags
+    ImageTag = apps.get_model('account', 'ImageTag')
+    windows_tag = ImageTag.objects.create(description='windows')
+    ImageTag.objects.create(description='rhel')
+
+    # Get current machine images
+    MachineImage = apps.get_model('account', 'MachineImage')
+    all_images = MachineImage.objects.filter(is_windows=True)
```
infinitewarp: This is a minor nitpick, but for readability I think `all_images` here should be renamed to something like `windows_images`. When I first looked at this function and tried to understand what it was doing, my lizard brain kept getting caught on "all" and I thought you were putting the windows tags on every image until I carefully reread each line.

kholdaway: 👍

cloudigrade/account/tasks.py
kholdaway: Yep.  Good catch.

cloudigrade/account/tasks.py
```diff
@@ -228,3 +235,55 @@ def run_inspection_cluster(messages, cloud='aws'):
...
+def persist_aws_inspection_cluster_results(inspection_result):
+    """
+    Persist the aws houndigrade inspection result.
+
+    Args:
+        inspection_result (dict): A dict containing houndigrade results
+    Returns:
+        None
+    """
+    RHEL_TAG = ImageTag.objects.filter(description='rhel').first()
```
infinitewarp: I think I understand why you capitalized this variable, but it still feels weird to me. The value being assigned to `RHEL_TAG` will _generally_ never change, which makes it _kind of_ like a constant, even though it has to be fetched each time and likely changes across environments. For consistency (and since `windows_tag` is lower-case elsewhere), I think this should be lower-cased so it looks like a regular variable.

That said, this does give me an idea to consider for the future but that we _don't_ need today. Since _within the same environment_ tags will probably never change, perhaps in the future we should create a helper function to get the tag objects (instead of going directly to the model class) which could memoize/cache the lookup to reduce redundant lookups.

cloudigrade/account/tests/test_tasks.py
```diff
@@ -441,3 +444,147 @@ def test_run_inspection_cluster_with_too_many_instances(self, mock_boto3):
...
+        """Assert that rhel_images are tagged rhel."""
+        ami_id = util_helper.generate_dummy_image_id()
+        user1 = util_helper.generate_test_user()
+        account1 = helper.generate_aws_account(user=user1)
+        machine_image1 = \
+            helper.generate_aws_image(account=account1,
+                                      is_encrypted=False,
+                                      is_windows=False,
+                                      ec2_ami_id=ami_id)
+        inspection_results = {'cloud': 'aws',
```
infinitewarp: Yikes… that indentation. 😯

While this is all technically valid, I wonder if reformatting it would improve readability? This version still fits in the line width rules:

```
        inspection_results = {
            'cloud': 'aws',
            'results': {
                ami_id: {
                    'drive': {
                        'partition': {
                            'rhel_found': True,
                            'evidence': [
                                {
                                    'release_file': '/redhat-release',
                                    'release_file_contents': 'RHEL\n',
                                    'rhel_found': True,
                                }
                            ],
                        }
                    }
                }
            }
        }
```

cloudigrade/account/tests/test_tasks.py
```diff
@@ -441,3 +444,147 @@ def test_run_inspection_cluster_with_too_many_instances(self, mock_boto3):
...
+        ami_id = util_helper.generate_dummy_image_id()
+        user1 = util_helper.generate_test_user()
+        account1 = helper.generate_aws_account(user=user1)
+        machine_image1 = \
+            helper.generate_aws_image(account=account1,
+                                      is_encrypted=False,
+                                      is_windows=False,
+                                      ec2_ami_id=ami_id)
+
+        inspection_results = {'cloud': 'aws',
```
infinitewarp: Seconded with the formatting like the previous test…

cloudigrade/util/models.py
```diff
@@ -12,3 +12,10 @@ class BaseModel(PolymorphicModel):
     class Meta:
         abstract = True
         ordering = ('created_at',)
+
+
+class BaseModel(models.Model):
+    """Abstract model to add automatic created_at and updated_at fields."""
+
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
```
infinitewarp: I think `BaseModel` also needs to include the inner `class Meta` copied from `BasePolymorphicModel`.

kholdaway:
@ilyavw Do you still have changes. PR merged is blocked by you.

PR #330, opened 2018-06-06
Make cloudigrade-test log more verbose

kdelee:
It would be a boon to QE if the test env logs were more verbose so if there are test failures we can know more about what was going on. Pretty sure changing this setting will do that, please advise if that is not the case.


kdelee:
Ok I'll close this then and we can talk more about another more secure way to access log data from test runs.

PR #331, opened 2018-06-06
special instructions to help pycurl work on macos

infinitewarp:


infinitewarp:
In my experience so far, `pycurl` seems to misbehave on _most_ operating systems. Even in out Docker image we have to include special a environment variable for it to install correctly. 😠  I wonder if team members using Fedora will encounter problems later when trying to install it…

elyezer:
It has been a while that I haven't used pycurl so I don't remember from the top of my head. I will make a note to try installing it and report any issue found.

Issue #334, opened 2018-06-07
500 if register a second cloud account that points to same AWS account with instance running


kdelee:
## Summary

If I create a second cloud account with a second, unique, ARN that points to the same AWS account as a previous cloud account and there is an instance running that was also discovered previously, I get a 500 error.

## Steps to Reproduce

1. Create two roles that grant access to the same AWS account
1. Have an instance running in that account
2. Create a cloud account on cloudigrade with one
3. Attempt to create a second one with the second ARN after the first one is done creating

## Expected Result

If I am not allowed to have two ARNs pointing to the same AWS account, maybe we should just check the account # for uniqueness too and return a 400.

## Actual Result

- 500 server error when it starts trying to inspect the instance (because it still does this synchronously at the moment)

infinitewarp:
Based on a discussion with several team members in Slack, we're doing to adjust the definition of this such that cloudigrade allows only _one_ ARN per AWS account. If someone tries to register a second ARN that shares the same account with an ARN that was already registered, we will reject that second request with a 400 error indicating that we already have an ARN for that account.

infinitewarp:
Demo of bugfix: https://asciinema.org/a/186124

kdelee:
Thanks!

PR #336, opened 2018-06-07
forbid multiple cloud accounts from having same aws account id

infinitewarp:
https://github.com/cloudigrade/cloudigrade/issues/334


infinitewarp:
Demo https://asciinema.org/a/186124

PR #338, opened 2018-06-07
add missing celery route config for run_inspection_cluster

infinitewarp:


infinitewarp:
I really don't think this prefix mess should have been necessary because I expected Celery/Kombu to automatically add them everywhere, but it seems that the configured prefix is only automatically added to task queues. ¯\\\_(ツ)\_/¯

Issue #339, opened 2018-06-07
de-queued ready_volumes message are not ack'd


infinitewarp:
## Summary

When `scale_up_inspection_cluster` runs, it reads messages from the `ready_volumes` queue, but afterwards AWS SQS keeps those messages in a "Messages in Flight" state which means the message was not correctly ack'd. As a result, the message eventually **re-queues** itself (after a configured time in AWS) and will get picked up by a future task. This is bad!

## Steps to Reproduce

1. register a cloud account that has an instance using an unknown non-windows ami
1. wait until `scale_up_inspection_cluster` completes
1. observe the message counts for `ready_volumes` in the SQS web console

## Expected Result

- 0 messages queued and 0 messages in flight

## Actual Result

- non-zero messages in flight

infinitewarp:
We seem to be experiencing the same problem as https://github.com/celery/kombu/issues/746. It looks like https://github.com/celery/kombu/pull/817 would fix it, and when I locally `pip install`ed that branch to experiment, everything SQS-related in cloudigrade seems to start working correctly, but that PR is over six months old and no closer to being merged.

The team has discussed our options, and I'm going to look at "fixing" our issue by dropping Kombu entirely and making straight `boto3` calls. 🤷‍♂️

ilyavw:
I can verify that this bug is resolved.

PR #342, opened 2018-06-08
replace kombu with direct boto3 calls for sqs

infinitewarp:
https://github.com/cloudigrade/cloudigrade/issues/339

### DIFF
--- a/cloudigrade/account/util.py
+++ b/cloudigrade/account/util.py
@@ -1,9 +1,13 @@
 """Various utility functions for the account app."""
+import base64
 import collections
-from queue import Empty
+import logging
+import math
+import uuid
 
-import kombu
-from django.conf import settings
+import boto3
+import jsonpickle
+from botocore.exceptions import ClientError
 from django.utils import timezone
 from django.utils.translation import gettext as _
 from rest_framework.serializers import ValidationError
@@ -12,6 +16,11 @@ from account import AWS_PROVIDER_STRING
 from account.models import (AwsInstance, AwsInstanceEvent, AwsMachineImage,
                             ImageTag, InstanceEvent)
 from util.aws import is_instance_windows
+
+logger = logging.getLogger(__name__)
+
+SQS_SEND_BATCH_SIZE = 10  # boto3 supports sending up to 10 items.
+SQS_RECEIVE_BATCH_SIZE = 10  # boto3 supports receiving of up to 10 items.
 
 
 def create_initial_aws_instance_events(account, instances_data):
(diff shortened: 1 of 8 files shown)


cloudigrade/account/util.py
```diff
@@ -126,80 +135,134 @@ def generate_aws_ami_messages(instances_data, ami_list):
...
+        message (object): message to encode and wrap
+
+    Returns:
+        dict: structured entry for sending to send_message_batch
+
+    """
+    return {
+        'Id': str(uuid.uuid4()),
+        # Yes, the outgoing message uses MessageBody, not Body.
+        'MessageBody': base64.b64encode(
```
kholdaway: What is the reason for Base64 encoding a JSON payload?

cloudigrade/account/util.py
```diff
@@ -126,80 +135,134 @@ def generate_aws_ami_messages(instances_data, ami_list):
...
+
+    Args:
+        sqs_message (dict): object to unwrap and decode
+
+    Returns:
+        object: the unwrapped and decoded message object
+
+    """
+    return jsonpickle.decode(
+        base64.b64decode(
```
kholdaway: Do we need to encode/decode JSON?  I see a need for binary, but not clear why we are doing this.

cloudigrade/account/util.py
```diff
@@ -126,80 +135,134 @@ def generate_aws_ami_messages(instances_data, ami_list):
...
+            sqs.delete_message(
+                QueueUrl=queue_url,
+                ReceiptHandle=sqs_message['ReceiptHandle'],
+            )
+            messages.append(unwrapped)
+        except ClientError as e:
+            # I'm not sure exactly what exceptions could land here, but we
+            # probably should log them, stop attempting further deletes, and
+            # return what we have received (and thus deleted!) so far.
+            logger.exception(e)
```
kholdaway: I think it would be helpful to add context to the error log.  What queue had failures?

cloudigrade/account/util.py
```diff
@@ -126,80 +135,134 @@ def generate_aws_ami_messages(instances_data, ami_list):
...
-                declare=[message_queue]
-            )
+    wrapped_messages = [_sqs_wrap_message(message) for message in messages]
+    batch_count = math.ceil(len(messages) / SQS_SEND_BATCH_SIZE)
+
+    for batch_num in range(batch_count):
+        start_pos = batch_num * SQS_SEND_BATCH_SIZE
+        end_pos = start_pos + SQS_SEND_BATCH_SIZE - 1
+        batch = wrapped_messages[start_pos:end_pos]
+        sqs.send_message_batch(QueueUrl=queue_url, Entries=batch)
```
kholdaway: Should we log failures?  It looks like boto provides failures in the response body if they occur.

cloudigrade/account/util.py
```diff
@@ -126,80 +135,134 @@ def generate_aws_ami_messages(instances_data, ami_list):
...
-    message_queue = kombu.Queue(
-        queue_name,
-        exchange=exchange,
-        routing_key=queue_name
+    sqs = boto3.client('sqs')
+    try:
+        return sqs.get_queue_url(QueueName=queue_name)['QueueUrl']
+    except ClientError as e:
+        if e.response['Error']['Code'].endswith('.NonExistentQueue'):
+            return sqs.create_queue(QueueName=queue_name)['QueueUrl']
```
kholdaway: Double failure?  Maybe log.

cloudigrade/account/util.py
infinitewarp: Hmmm… good catch. I've gone through so many phases of experimenting on this replacement that I forgot exactly _why_ this was necessary. It seems to be an artifact from a previous attempted implementation, and after some local testing today, it looks to be unnecessary. I'll rip it out in a subsequent push.

cloudigrade/account/util.py
infinitewarp: The AWS SQS message API (at least through boto3) cannot take most raw Python objects. It only works well with strings, and something on the producer and consumer sides need to do the conversion.

I chose `jsonpickle` to do that job because is has slightly better support for common non-primitives like `datetime` objects than the `json` module in the standard Python lib (which only supports a _very_ limited number of primitive types).

```
>>> things = [None, '🎉', 1, 1., 1j, b'123', True, (1, 2), [1, 2], {1, 2, 3}, {'hey': 'arnold!'}, datetime.datetime.now()]
>>>
>>> for thing in things:
...     print(jsonpickle.decode(jsonpickle.encode(thing)))
...
None
🎉
1
1.0
1j
b'123'
True
(1, 2)
[1, 2]
{1, 2, 3}
{'hey': 'arnold!'}
2018-06-14 16:29:51.742549
>>>
>>> for thing in things:
...     try:
...         print(json.loads(json.dumps(thing)))
...     except TypeError:
...         print(f'** json could not dumps {thing}')
...
None
🎉
1
1.0
** json could not dumps 1j
** json could not dumps b'123'
True
[1, 2]
[1, 2]
** json could not dumps {1, 2, 3}
{'hey': 'arnold!'}
** json could not dumps 2018-06-14 16:29:51.742549
>>>
```

cloudigrade/account/util.py
infinitewarp: Good idea! I'm adding more error logging in the next commit.

cloudigrade/account/util.py
infinitewarp: This isn't a double failure. As weird as it may look, I think this is the best way for us to _ensure that a queue exists_ before cloudigrade can either read from or write to it. Frameworks like Kombu would do this for us (for reasonable queue backends, at least), but since we're using bare `boto3` calls and `boto3` itself doesn't auto-create queues when needed, we have to implement this pattern of manually checking on one, creating it if not present, and then returning its URL.

cloudigrade/account/util.py
infinitewarp: At this point, if we can't write to the queue, there must be some rather catastrophic failure, and I think it makes sense to let the exception raise to the top (where it should be logged anyway).

The only reason I am explicitly handling and _working around_ exceptions when **reading** from the queue is that if we have an exception on the `n`th message, we would completely lose the `< n` messages since they would have already been deleted but not yet passed into the following function/task/whatever. It seemed more necessary to me to gracefully recover in that case, even though in that case it would also be pretty unexpected and catastrophic for the queue to be raising errors.

infinitewarp:
@kholdaway I've pushed a few more changes that I think address your comments. Please take a look and let me know what you think!

And if you're interested in seeing the _other_ side of the process, the sibling PR for houndigrade is https://github.com/cloudigrade/houndigrade/pull/27

cloudigrade/account/util.py
kholdaway: sorry, the question was worded poorly.  I meant base64 encode JSON not convert JSON string to python.

cloudigrade/account/util.py
kholdaway: I meant that creating a queue could result in a failure inside the except.

kholdaway: How are catastrophic failures logged?  Do we provide context around what we were trying to do?

cloudigrade/account/util.py
kholdaway: When an exception is raised to the top and logged, do we have the context to know what we were trying to do?  I haven't played with SQS a ton, but I know Kafka and other messaging queues can have issues where the system is up not not fully working.  Kafka can have a topic corrupted while another is working.  My suggestion for logging was to ensure we know if there is a global outage verse a localized failure (just a single queue).

Issue #353, opened 2018-06-18
Pycurl library required to run celery.


ilyavw:
## Summary

In trying to run celery locally I'm running into an unrecoverable error.

```
[2018-06-18 14:44:07,102: INFO/MainProcess] celery@iwhite-OSX.local ready.
[2018-06-18 14:44:07,103: CRITICAL/MainProcess] Unrecoverable error: ImportError('The curl client requires the pycurl library.',)
Traceback (most recent call last):
  File "/Users/iwhite/workspace/cloudigrade/cloudigrade/env/lib/python3.6/site-packages/kombu/asynchronous/http/__init__.py", line 20, in get_client
    return hub._current_http_client
AttributeError: 'Hub' object has no attribute '_current_http_client'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/iwhite/workspace/cloudigrade/cloudigrade/env/lib/python3.6/site-packages/celery/worker/worker.py", line 205, in start
    self.blueprint.start(self)
  File "/Users/iwhite/workspace/cloudigrade/cloudigrade/env/lib/python3.6/site-packages/celery/bootsteps.py", line 119, in start
    step.start(parent)
  File "/Users/iwhite/workspace/cloudigrade/cloudigrade/env/lib/python3.6/site-packages/celery/bootsteps.py", line 369, in start
    return self.obj.start()
  File "/Users/iwhite/workspace/cloudigrade/cloudigrade/env/lib/python3.6/site-packages/celery/worker/consumer/consumer.py", line 322, in start
    blueprint.start(self)
  File "/Users/iwhite/workspace/cloudigrade/cloudigrade/env/lib/python3.6/site-packages/celery/bootsteps.py", line 119, in start
    step.start(parent)
  File "/Users/iwhite/workspace/cloudigrade/cloudigrade/env/lib/python3.6/site-packages/celery/worker/consumer/consumer.py", line 598, in start
    c.loop(*c.loop_args())
  File "/Users/iwhite/workspace/cloudigrade/cloudigrade/env/lib/python3.6/site-packages/celery/worker/loops.py", line 91, in asynloop
    next(loop)
  File "/Users/iwhite/workspace/cloudigrade/cloudigrade/env/lib/python3.6/site-packages/kombu/asynchronous/hub.py", line 291, in create_loop
    item()
[...]

PR #354, opened 2018-06-18
keep kombu and pycurl because celery needs them

infinitewarp:
This reverts commit 988dd1d1188cafbba3b8ed33f024c478020c0709.

This should resolve https://github.com/cloudigrade/cloudigrade/issues/353

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -13,3 +13,4 @@ django-rest-polymorphic==0.1.7
 django-celery-beat==1.1.1
 django-health-check==3.6.1
 djoser==1.1.5
+pycurl==7.43.0.1
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,7 @@ RUN useradd -r cloudigrade
 RUN yum install centos-release-scl -y \
     && yum-config-manager --enable centos-sclo-rh-testing \
     && yum install \
+        libcurl-devel \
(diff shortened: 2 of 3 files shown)


requirements/base.txt
```diff
@@ -13,3 +13,5 @@ django-rest-polymorphic==0.1.7
 django-celery-beat==1.1.1
 django-health-check==3.6.1
 djoser==1.1.5
+kombu==4.2.1
```
ilyavw: I see that kombu also returns, can we restore just pycurl, or must we bring kombu in with it too?

infinitewarp: Celery uses Kombu under the covers, which in turn uses pycurl. Celery's dependency chain should automatically pull in Kombu, but pycurl is an optional dependency that you have to explicitly install.

infinitewarp: That being said… Celery _should_ automatically get Kombu. I could take it out to let Celery just get whatever version it thinks it needs. What are you thoughts on that, @ilyavw ?

ilyavw: I think that would be good, maybe one day it'll pull a version in that doesn't require pycurl.

ilyavw: At the same time... we should probably be explicit in our versioning, ¯\\\_(ツ)\_/¯

PR #359, opened 2018-06-19
Fix configuration for persist_inspection_cluster_results celery beat

### LINKED ISSUE #358: persist_inspection_cluster_results Celery beat requires CELERY_TASK_ROUTES entry to run
## Summary

As configured today, the persist_inspection_cluster_results celery beat will not run.  Even though persist_inspection_cluster_results uses a traditional messaging queue, the beat requires a route in the CELERY_TASK_ROUTES dict.

## Steps to Reproduce

1. Start Cloudigrade and notice the beat does not run.

## Expected Result
Beat should run.

## Actual Result
Beat doesn't run.

Fixing this bug will require a change to both cloudigrade and shiftigrade (2 PRs)

kholdaway:
Address issues found in #358

### DIFF
--- a/cloudigrade/account/tasks.py
+++ b/cloudigrade/account/tasks.py
@@ -213,6 +213,18 @@ def run_inspection_cluster(messages, cloud='aws'):
                 'command': task_command,
                 'environment': [
                     {
+                        'name': 'AWS_DEFAULT_REGION',
+                        'value': settings.AWS_SQS_REGION
+                    },
+                    {
+                        'name': 'AWS_ACCESS_KEY_ID',
+                        'value': settings.AWS_SQS_ACCESS_KEY_ID
+                    },
+                    {
+                        'name': 'AWS_SECRET_ACCESS_KEY',
+                        'value': settings.AWS_SQS_SECRET_ACCESS_KEY
+                    },
+                    {
                         'name': 'RESULTS_QUEUE_NAME',
                         'value': settings.HOUNDIGRADE_RESULTS_QUEUE_NAME
                     },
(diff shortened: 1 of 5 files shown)


cloudigrade/config/settings/base.py
```diff
@@ -249,6 +249,8 @@
         {'queue': 'scale_up_inspection_cluster'},
     'account.tasks.run_inspection_cluster':
         {'queue': 'run_inspection_cluster'},
+    'account.tasks.persist_inspection_cluster_results_task':
+        {'queue': 'persist_inspection_task'},
```
infinitewarp: So that we're consistent with all the naming pattern, let's change the queue name to match the function name `persist_inspection_cluster_results_task`.

infinitewarp: …and this will require another change to shiftigrade since you already merged that.

kholdaway: Pushed a change.  Thanks.

cloudigrade/account/tasks.py
```diff
@@ -279,6 +279,8 @@ def persist_inspection_cluster_results_task():
     messages = read_messages_from_queue(
         settings.HOUNDIGRADE_RESULTS_QUEUE_NAME,
         HOUNDIGRADE_MESSAGE_READ_LEN)
+    logger.info(_('{} read {} message(s) for processing'.format(
+        'persist_inspection_cluster_results_task', len(messages))))
```
infinitewarp: The parentheses here are incorrectly nested. I suspect you want it like:
```
logger.info(
   _('{} read {} message(s) for processing').format(
        'persist_inspection_cluster_results_task', len(messages)
    )
)
```

ilyavw: I would also recommend adding positional arguments here, in case we ever need to internationalize this string and the order of variables can change.

kholdaway: @ilyavw and @infinitewarp addressed both comments.

PR #368, opened 2018-06-21
Update houndigrade docs with availability zone

kdelee:
Include command to set availability zone from aws cli


kholdaway:
This PR doesn't cover the complete instructions so I will add another PR before closing issue.

Issue #357, opened 2018-06-18
copy_ami_snapshot fails for private/shared AMIs


infinitewarp:
## Summary

The `copy_ami_snapshot` task fails at the `add_snapshot_ownership` call for `boto3` when the target AMI is shared with _but not owned by_ the target AWS account. This results in the snapshot failing to copy and the whole pipeline of async tasks failing.

## Steps to Reproduce

1. Register a cloud account that has a running instance based on a _private shared_ AMI that _is not_ owned by the target account. 
    - `ami-065d0be9883aa064a` is an example that belongs to our test environment's AWS account that has been _shared_ with the various individual developer "customer" AWS accounts.
1. Run Celery and watch for the processing of `copy_ami_snapshot`.

## Expected Result

- Task runs and completes normally.

## Actual Result

- Task fails and produces errors that look like:
```
[2018-06-18 23:03:04,358: INFO/MainProcess] Received task: account.tasks.copy_ami_snapshot[7a82f697-69d8-4ff0-ae6f-60e97b6bcf01]
[2018-06-18 23:03:04,421: INFO/ForkPoolWorker-9] Found credentials in environment variables.
[2018-06-18 23:03:04,443: INFO/ForkPoolWorker-9] Starting new HTTPS connection (1): sts.amazonaws.com
[2018-06-18 23:03:04,874: INFO/ForkPoolWorker-9] Starting new HTTPS connection (1): ec2.us-east-1.amazonaws.com
[2018-06-18 23:03:05,142: INFO/ForkPoolWorker-9] Starting new HTTPS connection (1): ec2.us-east-1.amazonaws.com
[2018-06-18 23:03:05,369: INFO/ForkPoolWorker-9] Starting new HTTPS connection (1): sts.amazonaws.com
[2018-06-18 23:03:05,677: ERROR/ForkPoolWorker-9] Task account.tasks.copy_ami_snapshot[7a82f697-69d8-4ff0-ae6f-60e97b6bcf01] raised unexpected: RuntimeError("Unexpected AWS error <class 'botocore.exceptions.ClientError'>: An error occurred (InvalidSnapshot.NotFound) when calling the ModifySnapshotAttribute operation: None",)
Traceback (most recent call last):
  File "/Users/brasmith/projects/cloudigrade/cloudigrade/util/aws/helper.py", line 152, in wrapped
    result = original_function(*args, **kwargs)
[...]

PR #372, opened 2018-06-22
handle shared private images

infinitewarp:
https://github.com/cloudigrade/cloudigrade/issues/357

Demo: https://asciinema.org/a/189289

Also this will require a shiftigrade update for Celery to handle the new async task.


cloudigrade/account/tasks.py
```diff
@@ -67,10 +89,47 @@ def copy_ami_snapshot(arn, ami_id, snapshot_region):
...
+    if reference_ami_id is not None:
+        # If a reference ami exists, that means we have been working with a
+        # copy in here. That means we need to remove that copy and pass the
+        # original reference AMI ID through the rest of the task chain so the
+        # results get reported for that original reference AMI, not our copy.
+        # TODO FIXME Do we or don't we clean up?
+        # If we do, we need permissions to include `ec2:DeregisterImage` and
+        # `ec2:DeleteSnapshot` but those are both somewhat scary...
+        ami_id = reference_ami_id
+
```
abaiken: while I realize that we aren't cleaning up after ourselves at this time, are we going to somehow alert the customer to let them know that we created this copy in their account that they probably want to delete?

cloudigrade/util/aws/ec2.py
```diff
@@ -105,7 +103,68 @@ def get_ami(session, image_id, source_region):
...
+        image_id (str): An AMI ID
+        source_region (str): The region the snapshot resides in
+
+    Returns:
+        str: The image id of the newly created image.
+
+    """
+    old_image = get_ami(session, image_id, source_region)
+    # Note: AWS image names are limited to 128 characters in length.
+    new_name = f'cloudigrade reference copy ({old_image.name})'[:128]
```
abaiken: Not a big deal at all because I actually prefer to format strings this way, but do we want to move this to `' '.format()` to keep it more consistent with the string format we are moving towards?

cloudigrade/util/aws/helper.py
```diff
@@ -10,12 +11,14 @@
...
-# AWS has undocumented validation on the SnapshotId field in
-# snapshot related EC2 API calls.
+# AWS has undocumented validation on the Snapshot and Image Ids
+# when executing some "dryrun" commands.
 # We are uncertain what the pattern is. Manual testing revealed
 # that some codes pass and some fail, so for the time being
-# the value is hard-coded.
-SNAPSHOT_ID = 'snap-0f423c31dd96866b2'
+# the values are hard-coded.
+DRYRUN_SNAPSHOT_ID = 'snap-0f423c31dd96866b2'
```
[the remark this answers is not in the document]
ilyavw: Nope, that just needs to be a valid id. This goes back to validating policy actions, and ill-documented boto3 apis.

cloudigrade/util/aws/ec2.py
```diff
@@ -105,7 +103,68 @@ def get_ami(session, image_id, source_region):
...
+    """Raise an exception if image state is not available."""
+    if image.state == 'available':
+        return
+    message = _('Image {id} has state {state} (reason: {reason})').format(
+        id=image.id,
+        state=image.state,
+        reason=image.state_reason
+    )
+    if image.state == 'failed':
+        raise AwsImageError(message)
```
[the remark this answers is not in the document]
infinitewarp: `AwsImageError` does not extend `NotReadyException`, and that means that it should _not_ trigger the retry condition in the task decorator a few levels up. So, the task will report this exception and stop processing without retry.

The only case in which the task will retry (with backoff) is when the image state is _neither_ `'available'` _nor_ `'failed'`, which AFAICT are only the two terminal states.

cloudigrade/account/tasks.py
```diff
@@ -29,14 +29,19 @@
 
 @retriable_shared_task
 @rewrap_aws_errors
-def copy_ami_snapshot(arn, ami_id, snapshot_region):
+def copy_ami_snapshot(arn, ami_id, snapshot_region, reference_ami_id=None):
```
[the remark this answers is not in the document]
infinitewarp: Nope. The `reference_ami_id` is _only_ passed in specifically when the intent is to be aware of a reference/parent/original AMI. That's why it's an optional argument that defaults to `None`.

Having an optional argument like this is certainly not an _ideal_ solution in my eyes, but it's a compromise that that works here in lieu of either duplicating a _lot_ of code or refactoring it significantly for partial reuse among multiple different but similar implementations.

cloudigrade/util/aws/ec2.py
infinitewarp: Good point. I will change this to be more consistent.

cloudigrade/account/tasks.py
infinitewarp: True… we did say we need to capture some information that can eventually be given back to the end user. I will add a new model and some additional logic around it so we can do this.

infinitewarp:
Demo: https://asciinema.org/a/189289

Issue #148, opened 2018-03-21
Make the Log Analyzer scan periodically


infinitewarp:
As a Doppler admin, I want the log analyzing code to execute on its own so that I don't have to poke it manually.

## Acceptance Criteria 

- [x] Verify that there exists a periodic scheduled Celery task with a configured time (default: 2 minutes?) to do this work.
- [x] Verify that when a message is found on the SQS queue for CloudTrail, the corresponding S3 file is read and relevant information about all referenced instances, events, and images is stored to the database.
- [x] Verify that when a "start" event is found for an instance using an image we have already seen, we do not start the inspection process for that image.
- [x] Verify that when a "start" event is found for an instance using an image we've never seen and that image has the "windows" platform set, the image is saved with the `is_windows` property with an "inspected" status.
- [x] Verify that when a "start" event is found for an instance using an image we've never seen and the "windows" platform is _not_ set, the `copy_snapshot` async task is called for start the inspection process.
- ~[ ] Verify that an integration test exists for this functionality.~
  - ~[ ] Verify that issues have been defined in the integrade repository for tests.~
  - We've decided not to add integration tests. See discussion comments below…

## Assumptions and Questions

- We need to refactor out the existing Django code that @adberglund wrote in previous story. Management commands were deleted in commit `37be5709a6a2f5f377d64364c06368e84599b86c`, the file was `analyze_log.py`. See also Wacky Dangler issue https://github.com/cloudigrade/cloudigrade/issues/114.
- We acknowledge that running this as a periodic scheduled task is not ideal compared to a daemon that is constantly listening to the queue. We accept this architectural compromise to keep the system simple and quick to develop for now. We may revisit this topic later, especially if or when we need to address performance bottlenecks or timing issues.

abaiken:
I have verified all of the AC's except for integration tests via Ilya's [demo](https://asciinema.org/a/TVS3j7pndjqIFvQTJtDLUEGsQ). @elyezer or @kdelee could one of you possibly create issues on integrade for the log analyzer and then verify the last AC to close the issue?

elyezer:
Sure, I am working on that right now.

elyezer:
There is not external API that we can perform integration tests for this issue. We will be able to test this when the reports API are done so that we can create usage on the fly and check if the Log Analyzer has really inspected as it should.

Let me know if you have any idea on how to test this before the reporting being done.

infinitewarp:
[...]

PR #389, opened 2018-06-30
Tag openshift when account is added

kholdaway:
This PR handles the initial tagging of AMIs when an account is added.  I figure I can break this into two pieces.  

1.  Initial tagging when account is added (this PR)
2. Tagging with cloud trails events.  This depends on https://github.com/cloudigrade/cloudigrade/issues/148.  I do this in a separate PR.


cloudigrade/account/migrations/0010_openshift_tag.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.0.6 on 2018-06-28 18:46
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+def add_openshift_tag(apps, schema_editor):
+    """Copy old is_windows value to an image tag."""
```
abaiken: Remove `is_windows` from docstring.

cloudigrade/account/serializers.py
```diff
@@ -129,6 +134,27 @@ def create(self, validated_data):
             )
         return account
 
+    def add_openshift_tag(self, session, ami_id, ami_region, image):
```
abaiken: Not a blocker, but possibly rename this function since we aren't always adding the openshift tag? Possibly `check_for_openshift` or something along those lines?

PR #391, opened 2018-07-02
148 log analyzer

ilyavw:
#148 

Demo: https://asciinema.org/a/TVS3j7pndjqIFvQTJtDLUEGsQ

### DIFF
--- a/cloudigrade/account/util.py
+++ b/cloudigrade/account/util.py
@@ -45,22 +45,55 @@ def create_initial_aws_instance_events(account, instances_data):
     saved_instances = collections.defaultdict(list)
     for region, instances in instances_data.items():
         for instance_data in instances:
-            instance, __ = AwsInstance.objects.get_or_create(
-                account=account,
-                ec2_instance_id=instance_data['InstanceId'],
-                region=region,
-            )
+            saved_instances[region].append(
+                save_instance_events(account, instance_data, region))
+    return dict(saved_instances)
+
+
+def save_instance_events(account, instance_data, region, events=None):
+    """
+    Save provided events, and create the instance object if it does not exist.
+
+    Args:
+        account (AwsAccount): The account that owns the instance that spawned
+            the data for these InstanceEvents.
+        instance_data (dict): Dictionary containing instance information.
+        region (str): AWS Region.
+        events (list): List of Events to be saved.
+
+    Returns:
+        AwsInstance: Object representing the saved instance.
+
+    """
+    instance, __ = AwsInstance.objects.get_or_create(
+        account=account,
+        ec2_instance_id=instance_data['InstanceId'] if isinstance(
+            instance_data, dict) else instance_data.instance_id,
+        region=region,
+    )
+    if events is None:
+        # Assume this is the initial event
+        event = AwsInstanceEvent(
+            instance=instance,
+            event_type=InstanceEvent.TYPE.power_on,
+            occurred_at=timezone.now(),
+            subnet=instance_data['SubnetId'],
+            ec2_ami_id=instance_data['ImageId'],
+            instance_type=instance_data['InstanceType'],
+        )
+        event.save()
+    else:
+        for event in events:
             event = AwsInstanceEvent(
                 instance=instance,
-                event_type=InstanceEvent.TYPE.power_on,
-                occurred_at=timezone.now(),
-                subnet=instance_data['SubnetId'],
-                ec2_ami_id=instance_data['ImageId'],
-                instance_type=instance_data['InstanceType'],
+                event_type=event['event_type'],
+                occurred_at=event['occurred_at'],
+                subnet=event['subnet'],
+                ec2_ami_id=event['ec2_ami_id'],
+                instance_type=event['instance_type'],
             )
             event.save()
-            saved_instances[region].append(instance)
-    return dict(saved_instances)
+    return instance
 
 
 def create_new_machine_images(account, instances_data):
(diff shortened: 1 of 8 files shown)


cloudigrade/account/util.py
```diff
@@ -88,22 +121,71 @@ def create_new_machine_images(account, instances_data):
...
+    """
+    Start image inspection of the provided image.
+
+    Args:
+        arn (str):  The AWS Resource Number for the account with the snapshot
+        image_id (str): The AWS ID for the machine image
+        region (str): The region the snapshot resides in
+
+    """
+    image = AwsMachineImage.objects.get(ec2_ami_id=image_id)
```
infinitewarp: In this function, you call the object `image` but in the preceding ones you call it `ami`. Let's pick one to use consistently through this change, preferably `ami` since that's more specific.

cloudigrade/account/util.py
```diff
@@ -88,22 +121,71 @@ def create_new_machine_images(account, instances_data):
...
+    image.status = image.PREPARING
+    image.save()
+
+    from account.tasks import copy_ami_snapshot
```
infinitewarp: Why are you importing this locally instead of at the start of the module? Is there a potential circular import that I've overlooked? If so, throw a comment in here to explain so that doesn't accidentally get refactored out.

cloudigrade/analyzer/tasks.py
```diff
@@ -0,0 +1,176 @@
...
+
+    aws.delete_message_from_queue(queue_url, messages)
+
+
+def _parse_log_for_ec2_events(log):
+    """
+    Parse S3 log for EC2 on/off events.
+
+    Args:
+        log (string): The string contents of the log file.
```
infinitewarp: `str` ≠ `string` 🙂

cloudigrade/analyzer/tasks.py
```diff
@@ -0,0 +1,176 @@
...
+
+    Returns:
+        list(tuple): List of instance data seen in log.
+        list(tuple): List of instance_event data seen in log.
+
+    """
+    ec2_event_map: Dict[str, Any] = {
```
infinitewarp: I wonder if this should be moved up in the module as a pseudo-constant since it doesn't need to be redefined on every function call.

infinitewarp: Also, interesting use of type hinting! I haven't checked, but I think this might be the first type hint in cloudigrade. What was your rationale for using it here?

cloudigrade/account/util.py
```diff
@@ -88,22 +121,71 @@ def create_new_machine_images(account, instances_data):
...
+
+    copy_ami_snapshot.delay(str(arn), image_id, region)
```
abaiken: If the arn is coming in as a string, is there any reason to cast it as a string here?

cloudigrade/util/aws/ec2.py
```diff
@@ -384,4 +384,6 @@ def is_instance_windows(instance_data):
         bool: True if it appears to be windows, else False.
 
     """
-    return instance_data.get('Platform', '').lower() == 'windows'
+    return instance_data.get('Platform', '').lower() == 'windows' if \
+        isinstance(instance_data, dict) \
```
infinitewarp: In what specific cases will `instance_data` not be a `dict`? If we intend to play polymorphically here and handle multiple object types, we should update the docstring above.

cloudigrade/analyzer/tasks.py
infinitewarp: Or to be more clear: I think you meant to use the `str` type here because `string` is a module.

cloudigrade/config/settings/base.py
```diff
@@ -234,6 +234,11 @@
                                       default=AWS_NAME_PREFIX + \
                                               'inspection_results')
 
+CLOUDTRAIL_EVENT_URL = env(
```
abaiken: should we document what to set this environment variable to somewhere?

cloudigrade/analyzer/tasks.py
ilyavw: The type hinting was a left over from the code originally written by Andrew, upon a second look we don't need it, so it's gone 🎉

cloudigrade/util/aws/ec2.py
ilyavw: Updated the doc with more info on this.

cloudigrade/config/settings/base.py
ilyavw: Cloudigrade and Shiftigrade readmes have been updated.

ilyavw:
@infinitewarp @abaiken Change made, let me know what you think!

cloudigrade/account/util.py
```diff
@@ -45,22 +45,55 @@ def create_initial_aws_instance_events(account, instances_data):
...
-                account=account,
-                ec2_instance_id=instance_data['InstanceId'],
-                region=region,
-            )
+            saved_instances[region].append(
+                save_instance_events(account, instance_data, region))
+    return dict(saved_instances)
+
+
+def save_instance_events(account, instance_data, region, events=None):
```
kholdaway: Should we put AWS in the function name? Just looking forward to Azure. This method clearly expects AWS.  Or we punt this to another issue to make sure we handle things more generally (when we do Azure work).  Not a blocker.

ilyavw: I had the same thought, but I figure we punt this until we are actually implementing multi-cloud as it'll require a lot of other changes at the same time. A lot of these might get refactored to `util.aws` at that time.

cloudigrade/account/util.py
```diff
@@ -88,22 +121,72 @@ def create_new_machine_images(account, instances_data):
...
 
+def save_machine_images(account, ami_id):
+    """
+    Save the image object.
+
+    Args:
+        account (AwsAccount): The account associated with the machine image
+        ami_id (str): The AWS AMI ID.
+
+    Returns:
```
infinitewarp: Missing return docstring.

PR #397, opened 2018-07-03
Check if image tags attribute is None

kholdaway:


cloudigrade/account/serializers.py
```diff
@@ -142,11 +142,12 @@ def add_openshift_tag(self, session, ami_id, ami_region, image):
 
         """
         ec2_image = aws.get_ami(session, ami_id, ami_region)
-        has_openshift = 'cloudigrade-ocp-present' in [
-            tags['Key'] for tags in ec2_image.tags]
-        if has_openshift:
-            image.tags.add(ImageTag.objects.filter(
-                description='openshift').first())
+        if ec2_image.tags is not None:
```
kdelee: So it is either `None` or is a dictionary with the key `'Key'` all the time?

kholdaway: Yep. Unless amazon changes their APi.

cloudigrade/account/tests/test_serializers.py
```diff
@@ -209,6 +209,39 @@ def test_openshift_tag_not_added(self, mock_aws, mock_check_image_state):
         openshift_tag = test_image.tags.filter(description='openshift').first()
         self.assertEqual(openshift_tag, None)
 
+    @patch('util.aws.ec2.check_image_state')
+    @patch('account.tasks.aws')
+    def test_openshift_no_tags(self, mock_aws, mock_check_image_state):
+        """Test case where AMI has no tagsblue."""
```
abaiken: Is the docstring supposed to say " no tagsblue."

infinitewarp:
It looks like this should solve the problem I was seeing. Normally I would wait and let the author merge, but since this is a blocker, I'm going to go ahead and merge this now.

PR #402, opened 2018-07-06
Add docs for automated aws setup of cloudigrade resources

kdelee:
@ironfroggy here is a prettier version of notes from my issue

In summary, @ironfroggy tried this out today and was able to create/tear down the houndigrade cluster and s3 setup. We will probably want to work on migrating this code to a cloudigrade namepace and some tweaks may be desirable for gitlab CI, but both of these things are probably better done after the migration. So this PR is just to document the usage, and we can update the paths to the roles when they get moved.

I did notice there is no mention of any of these docs in the README, I could change that if desirable, or we could keep it the way it is.

### DIFF
--- a/docs/dev-automated-aws-setup.md
+++ b/docs/dev-automated-aws-setup.md
@@ -1,0 +1,130 @@
+# Cloudigrade AWS setup with Ansible
+
+This is an alternative, automated workflow to the manual workflows described in [dev-houndigrade-cluster-setup.md](https://github.com/cloudigrade/cloudigrade/blob/master/docs/dev-houndigrade-cluster-setup.md) and [dev-cloudtrail-initial-setup.rst](https://github.com/cloudigrade/cloudigrade/blob/master/docs/dev-cloudtrail-initial-setup.rst). They are mutually exclusive.
+
+## Assumptions + Dependencies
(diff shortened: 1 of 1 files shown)


docs/dev-automated-aws-setup.md
```diff
@@ -0,0 +1,138 @@
+# Cloudigrade AWS setup with Ansible
+
+This is an alternative, automated workflow to the manual workflows described in [dev-houndigrade-cluster-setup.md](https://github.com/cloudigrade/cloudigrade/blob/master/docs/dev-houndigrade-cluster-setup.md) and [dev-cloudtrail-initial-setup.rst](https://github.com/cloudigrade/cloudigrade/blob/master/docs/dev-cloudtrail-initial-setup.rst). They are mutually exclusive.
+
+## Assumptions + Dependencies
+
+* Python executable in path must have boto3 importable
```
ironfroggy: Is the version or virtual env of the Python executable important?

kdelee: I was on python 2.7 and you were on python 3.6 (pretty sure), so AFAIK it is not important. I'm sure there could be some weird edge case.

Issue #356, opened 2018-06-18
API to get summary totals per cloud account


infinitewarp:
## Summary

As a Cloud Meter API client, I want to see a list of my cloud accounts ("ARNs") with each having 1) total number of found images, 2) total number of found instances, 3) total number of instances identified as using a RHEL image, 4) total number of instances identified as using an OpenShift image, all of which are filtered to a defined date range.

This will power the rows in the lower half of the dashboard page. See [this mockup](https://redhat.invisionapp.com/share/UFI078IZVAW#/screens/296433286).

Inputs for the API call must include the start date and end date. These are REQUIRED arguments. The API will not restrict the allowed range of days. Leaving this open allows us to find performance issues to optimize in the future.

Input may include OPTIONAL `user_id` argument to facilitate admin/superuser functionality. This should behave like our other APIs. Normal users will not supply this value and already get filtered by presense of session. Superusers will supply this value to impersonate normal users.

Output data is effectively aggregated by cloud account. For each cloud account, output includes:

- Number of distinct images used for _any_ instance seen running during the period.
- Number of distinct instances seen running during the period.
- Number of distinct instances with a RHEL-tagged image seen running during the period.
- Number of distinct instances with an OpenShift-tagged image seen running during the period.

Note: It does not matter if an instance starts and stop multiple times during the period. The instance only contributes to the counts exactly once if ever seen.

Output should look something like this:

```json
{
    "cloud_account_overviews": [
        {
            "id": 13,
            "user_id": 4,
            "type": "aws",
            "arn": "arn:aws:iam::1234567890:role/rolename",
            "creation_date": "2018-01-15T12:34:56Z",
            "name": "my great account",
            "images": 10,
            "instances": 50,
            "rhel_instances": 30,
            "openshift_instances": 30
        },
        {
            "id": 14,
            "user_id": 4,
            "type": "aws",
            "arn": "arn:aws:iam::1234567891:role/rolename",
            "creation_date": "2018-02-07T12:34:56Z",
            "name": "my other account",
            "images": 10,
            "instances": 50,
            "rhel_instances": 30,
            "openshift_instances": 30
        },
    ]
}
```

If we have not yet built the ability for a user to define a custom display name for their cloud accounts, we should still include a `name` attribute but leave it `None`/`null` for now.

The API should live at: `/api/v1/report/accounts/`

## Acceptance Criteria

- [x] Verify that "images" represents the total number of unique images for instances that were powered on during the reporting period.
[...]

PR #403, opened 2018-07-06
API for totals per clount (cloud account overview)

abaiken:
Issue: #356 
Demo: Coming soon


cloudigrade/account/reports/__init__.py
```diff
@@ -179,3 +187,113 @@ def _calculate_instance_usage(start, end, events):
...
+
+    Args:
+        start (datetime.datetime): Start time (inclusive)
+        end (datetime.datetime): End time (exclusive)
+        account (Account or AwsAccount): Account or AwsAccount object
+
+    Returns:
+        dict: An overview of the instances/images/rhel & openshift images for the specified
+        account during the specified time period."""
+    machine_images = AwsMachineImage.objects.all()
```
infinitewarp: This is redundant and risks introducing performance issues by fetching _all_ images. You could simply remove this line and change your filter below...

cloudigrade/account/reports/__init__.py
```diff
@@ -179,3 +187,113 @@ def _calculate_instance_usage(start, end, events):
...
+        cloud_helper = helper.get_report_helper(AWS_PROVIDER_STRING, account.aws_account_id)
+        # _get_relevant_events will return the events in between the start & end times & if
+        # no events are present during this period, it will return the last event that occurred
+        events = _get_relevant_events(start, end, cloud_helper)
+        for event in events:
+            valid_event = validate_event(event, start)
+            if valid_event:
+                instances.append(event.instance.id)
+                images.append(event.ec2_ami_id)
+                image = machine_images.filter(ec2_ami_id=event.ec2_ami_id).first()
```
infinitewarp: ...to use the model's objects directly like `AwsMachineImage.objects.filter(...).first()`.

Or even better, you can simplify that to `AwsMachineImage.objects.get(...)` to get exactly the one match. If no match is found, `get` will raise an exception, but that's okay here because we really _should_ have that image saved at this point.

(If we merge cloudigrade/pull/405, this line can go away entirely since you will be able to get the image at `event.machineimage`.)

cloudigrade/account/reports/__init__.py
```diff
@@ -179,3 +187,113 @@ def _calculate_instance_usage(start, end, events):
...
+                if image:
```
infinitewarp: Following up on the previous comment, is there ever a case where we wouldn't have an image saved for the event? I don't think there is, and if there isn't, we can lose this condition.

(This will be especially true if we merge cloudigrade/pull/405 which will _require_ the image to exist in this relationship.)

cloudigrade/account/reports/__init__.py
```diff
@@ -179,3 +187,113 @@ def _calculate_instance_usage(start, end, events):
...
+                    for tag in image.tags.all():
+                        if tag.description == 'rhel':
+                            rhel.append(image)
+                        if tag.description == 'openshift':
+                            openshift.append(image)
+        # set aws account info & grab the totals
+        cloud_type = AWS_PROVIDER_STRING
+        arn = account.account_arn
+        id = account.aws_account_id
```
infinitewarp: `id` is a built-in function in Python, and I'm surprised `flake8` didn't complain about overwriting it. Regardless, we should name this variable something else so we don't overwrite `id`.

cloudigrade/account/reports/__init__.py
```diff
@@ -179,3 +187,113 @@ def _calculate_instance_usage(start, end, events):
...
+        account during the specified time period."""
+    machine_images = AwsMachineImage.objects.all()
+    instances = []
+    images = []
+    rhel = []
+    openshift = []
+    cloud_type = 'unknown'
+    arn = 'unsupported'
+    # make sure that this is a supported cloud type (aws at this time)
+    if isinstance(account, AwsAccount):
```
infinitewarp: I think we can lose this condition and all of the `else` block. _All_ of our current accounts are AWS, which means de facto all account types are supported for reporting. When we add new account types, those will need to get exposed in the same way, and I expect we will require them to have a similar-enough class hierarchy to reuse throughout our reporting/dashboard APIs.

(If we merge cloudigrade/pull/405, we should be in a place to make that happen since we can stop referring to the AWS-specific `ec2_ami_id` and just look at the "base" `MachineImage` related to the "base" `InstanceEvent`.)

cloudigrade/account/reports/__init__.py
```diff
@@ -179,3 +187,113 @@ def _calculate_instance_usage(start, end, events):
...
+        cloud_helper = helper.get_report_helper(AWS_PROVIDER_STRING, account.aws_account_id)
+        # _get_relevant_events will return the events in between the start & end times & if
+        # no events are present during this period, it will return the last event that occurred
+        events = _get_relevant_events(start, end, cloud_helper)
+        for event in events:
+            valid_event = validate_event(event, start)
+            if valid_event:
+                instances.append(event.instance.id)
+                images.append(event.ec2_ami_id)
```
infinitewarp: FWIW, if we merge cloudigrade/pull/405, this will need to change from `event.ec2_ami_id` to something like `event.machineimage.id`, which will make this nice and generic.

cloudigrade/account/reports/__init__.py
```diff
@@ -179,3 +187,113 @@ def _calculate_instance_usage(start, end, events):
...
+                image = machine_images.filter(ec2_ami_id=event.ec2_ami_id).first()
+                if image:
+                    for tag in image.tags.all():
+                        if tag.description == 'rhel':
+                            rhel.append(image)
```
infinitewarp: Since we only care about building a unique set later on, we could simply append the `image.id` to reduce the memory footprint of these lists.

cloudigrade/account/reports/__init__.py
```diff
@@ -179,3 +187,113 @@ def _calculate_instance_usage(start, end, events):
...
+        id = account.id
+        logger.warning(_('Account "{0}" has an unsupported cloud type "{1}", therefore'
+                         'there is no data on its images/instances during the specified'
+                         'start and end dates').format(account, cloud_type))
+        total_images, total_instances, total_rhel, total_openshift = None, None, None, None
+    # if the account was created right at or after the end time, we cannot give meaningful
+    # data about the instances/images seen during the period, therefore we need to make sure
+    # that we return None for those values
+    valid_account_data = validate_account_creation(account, end)
+    if not valid_account_data:
```
infinitewarp: I would put this condition at the very start of this function and let it return early if triggered. Since it just throws away the data that would have been collected, it should run _first_ and short-circuit the rest of the logic.

cloudigrade/account/reports/__init__.py
```diff
@@ -179,3 +187,113 @@ def _calculate_instance_usage(start, end, events):
...
+    """
+    valid_event = True
+    # if the event occurred outside of our specified period (ie. before start) we should
+    # only inspect it if it was a power on event
+    if event.occurred_at < start:
+        if event.event_type == InstanceEvent.TYPE.power_off:
+            valid_event = False
+    return valid_event
+
+def validate_account_creation(account, end):
```
infinitewarp: Does this really need to be its own function? It feel like overkill to replace a simple and readable one-liner (`if end > account.created_at`) with a potentially less obvious one (the word "validate" doesn't feel right to me here).

cloudigrade/account/reports/__init__.py
```diff
@@ -179,3 +187,113 @@ def _calculate_instance_usage(start, end, events):
...
+                images.append(event.ec2_ami_id)
+                image = machine_images.filter(ec2_ami_id=event.ec2_ami_id).first()
+                if image:
+                    for tag in image.tags.all():
+                        if tag.description == 'rhel':
+                            rhel.append(image)
+                        if tag.description == 'openshift':
+                            openshift.append(image)
+        # set aws account info & grab the totals
+        cloud_type = AWS_PROVIDER_STRING
```
infinitewarp: I suspect this isn't (or won't be) the only place we would directly reference `AWS_PROVIDER_STRING` in places we're dealing specifically with `AwsAccount` objects. Wouldn't it be nice if `Account` objects had a property that we could just look at instead of inferring based on things like the `isinstance(account, AwsAccount)` a few lines above this?

Now might be a good time to add that. If so, I would add an abstract property to the `Account` class and an actual property to the `AwsAccount` class that returns `AWS_PROVIDER_STRING` directly. Then, whenever we need to know the provider, we can just look at the instance like `account.cloud_provider`, and that's a lot cleaner than `isinstance` checking. If you're not sure how to do this, see `product_identifier` as a similar implementation on `InstanceEvent`/`AwsInstanceEvent`.

...or maybe my suggestion is overkill? I took a look, and there currently aren't many places `AWS_PROVIDER_STRING` is actually referenced, but we _will_ have to do something _like_ this when we expand beyond AWS. Food for thought.

cloudigrade/account/tests/helper.py
```diff
@@ -41,6 +42,29 @@ def generate_aws_account(arn=None, aws_account_id=None, user=None, name=None):
     )
 
 
+def generate_account(user=None, name=None):
```
infinitewarp: Although it's _technically_ allowed in the current state of the code, realistically nothing should be creating barren `Account` objects without also creating a cloud-specific account object (e.g. `AwsAccount`) that references/extends it, and when we add new cloud providers, they will fall under the same rule and make their own cloud-specific account objects (e.g. `AzureAccount`) to do the same.

`Account` isn't strictly abstract (maybe it _should_ be, but I suspect there were ORM/framework reasons for this that I don't recall immediately), but conceptually that's how we should treat it and the other "base" classes (`Instance`, `InstanceEvent`, and `MachineImage`).

Unless there's a strong reason for divorcing the base object from its cloud-specific object, I think we should be using the existing `generate_aws_account` helper instead of this.

cloudigrade/account/reports/__init__.py
abaiken: This makes sense! I ran into this in testing because I was using the `generate_aws_instance_events` helper function and didn't specify an `ec2_ami_id` and forgot to revisit it (fixed now)! Great comments!

cloudigrade/account/reports/__init__.py
abaiken: Good catch! Weird flake8 didn't catch this 🤷‍♀️

cloudigrade/account/reports/__init__.py
abaiken: I definitely agree that it would be nice to have the type property built into the accounts and we should plan on this for the future, but I don't think that we should add onto this issue to do it.

cloudigrade/account/views.py
```diff
@@ -150,3 +150,30 @@ def list(self, *args, **kwargs):
...
+        if user_id is not None:
+            user_id = convert_param_to_int('user_id', user_id)
+            accounts = accounts.filter(user__id=user_id)
+        overviews = {'cloud_account_overviews': []}
+        # iterate through each account in our queryset and append the
+        # overview to our cloud_account_overviews list
+        for account in accounts:
+            account_overview = serializer.get_overview(account)
+            overviews['cloud_account_overviews'].append(account_overview)
+        return Response(overviews, status=status.HTTP_200_OK)
```
infinitewarp: Is `status=status.HTTP_200_OK` necessary here? I think the default `status` for a `Response` is already `200` since it inherits (through several layers) from `django.http.response.HttpResponseBase`, but I haven't tested to confirm this.

cloudigrade/config/urls.py
```diff
@@ -18,6 +19,8 @@
 router.register(r'image', MachineImageViewSet)
 router.register(r'report', ReportViewSet, base_name='report')
 router.register(r'sysconfig', SysconfigViewSet, base_name='sysconfig')
+router.register(r'report/accounts', CloudAccountOverviewViewSet,
+                base_name='overview')
```
infinitewarp: Although not required, the convention for naming routes is to match the relevant object's class name (if it's a typical CRUD endpoint) or roughly match the URL path it's registered to. That makes it more obvious when debugging or if you need to manually `reverse` a URL later. For something like this, a name like `report-accounts` might be appropriate.

cloudigrade/account/reports/__init__.py
```diff
@@ -179,3 +187,82 @@ def _calculate_instance_usage(start, end, events):
...
+    return valid_event
+
+def get_account_overview(account, start, end):
+    """
+    Generate an overview of an account over a specified amount of time.
+
+    Args:
+        start (datetime.datetime): Start time (inclusive)
+        end (datetime.datetime): End time (exclusive)
+        account (AwsAccount): AwsAccount object
```
infinitewarp: The args in this docstring don't match the order of the args in the function signature.

cloudigrade/account/tests/test_reports_aws.py
```diff
@@ -342,3 +343,356 @@ def test_usage_on_times_overlapping(self):
...
+             'creation_date': self.account.created_at,
+             'name': self.account.name,
+             'images': 2,
+             'instances': 2,
+             'rhel_instances': 2,
+             'openshift_instances': 1}
+        self.assertEqual(expected_overview, overview)
+
+    def test_get_cloud_account_overview_with_openshift(self):
+        """Assert an account overview with events reports correctly."""
```
infinitewarp: This docstring is identical to the ones for several preceding test functions. They should have enough information to distinguish different tests.

cloudigrade/account/tests/test_views.py
```diff
@@ -944,3 +946,212 @@ def test_list_accounts_success(self, mock_get_primary_account_id):
...
+             'user_id': self.account3.user_id,
+             'type': 'aws',
+             'arn': self.account3.account_arn,
+             'creation_date': self.account3.created_at,
+             'name': self.account3.name,
+             'images': 1,
+             'instances': 1,
+             'rhel_instances': 0,
+             'openshift_instances': 1}
+        self.account4_expected_overview = \
```
infinitewarp: This is a minor style comment, but if you put the open curly brace on this line, you don't need to escape this newline.

docs/rest-api-examples.rst
```diff
@@ -33,6 +33,7 @@ The following resource paths are currently available:
 
 -  ``/api/v1/account/`` returns account data
 -  ``/api/v1/report/`` returns usage report data
+-  ``/api/v1/report/accounts`` returns account overview data
```
infinitewarp: Trailing slash on the URL for consistency with the other examples.

cloudigrade/account/reports/__init__.py
```diff
@@ -179,3 +187,82 @@ def _calculate_instance_usage(start, end, events):
...
+        # no events are present during this period, it will return the last event that occurred
+        events = _get_relevant_events(start, end, cloud_helper)
+        for event in events:
+            valid_event = validate_event(event, start)
+            if valid_event:
+                instances.append(event.instance.id)
+                images.append(event.ec2_ami_id)
+                image = AwsMachineImage.objects.get(ec2_ami_id=event.ec2_ami_id)
+                for tag in image.tags.all():
+                    if tag.description == 'rhel':
```
infinitewarp: It's not something we need to fix _now_ with your commit here, but as I see we are using magic string tokens to indicate "rhel" and "openshift", I keep thinking that we should have a better way of recognizing these special tags. I'm just "thinking aloud" here but recording it so we don't forget…

Issue #355, opened 2018-06-18
API to get daily totals over date range


infinitewarp:
## Summary

As a Cloud Meter API client, I want to 1) see the total count of instances that were running a RHEL or OpenShift image for each day and 2) the total elapsed seconds they ran for each day in a given period so that I can build pretty graphs to show to users.

This will power the three line/bar graphs on the dashboard page. See [this mockup](https://redhat.invisionapp.com/share/UFI078IZVAW#/screens/296433286).

Inputs for the API call must include the start date and end date. These are REQUIRED arguments. The API will not restrict the allowed range of days. Leaving this open allows us to find performance issues to optimize in the future.

Input may include OPTIONAL `user_id` argument to facilitate admin/superuser functionality. This should behave like our other APIs. Normal users will not supply this value and already get filtered by presense of session. Superusers will supply this value to impersonate normal users.

Output data is effectively aggregated by day. For each day, output includes:

- Number of instances with a RHEL-tagged image _seen running_ during the day. It doesn't matter if the same instance started and stopped multiple times; we care about the number of distinct instances seen.
- Number of instances with an OpenShift-tagged image _seen running_ during the day. It doesn't matter if the same instance started and stopped multiple times; we care about the number of distinct instances seen.
- Cumulative number of seconds any instances with a RHEL-tagged image ran during the day.
- Cumulative number of seconds any instances with an OpenShift-tagged image ran during the day.

If an instance is backed by an image that has BOTH RHEL and OpenShift tags, it fairly contributes to both categories.

Output should look something like this:

```json
"data": [
  {
    "date": "2018-06-01",
    "rhel_instances": 12,
    "openshift_instances": 12,
    "rhel_runtime_seconds": 6000,
    "openshift_runtime_seconds": 6000
  },
  {
    "date": "2018-06-02",
    "rhel_instances": 12,
    "openshift_instances": 12,
    "rhel_runtime_seconds": 6000,
    "openshift_runtime_seconds": 6000
  },
  {
    "date": "2018-06-03",
    "rhel_instances": 12,
    "openshift_instances": 12,
    "rhel_runtime_seconds": 6000,
    "openshift_runtime_seconds": 6000
  }
]
```

The API should live at: `/api/v1/report/instances/`

## Acceptance Criteria

- For RHEL-tagged and OpenShift-tagged images:
    - Verify an instance that was started on day `x` and stopped on day `x+2`:
        - [ ] contributes once to the instance count for day `x`
[...]

PR #405, opened 2018-07-06
add instanceevent to machineimage relationship

infinitewarp:
Related to https://github.com/cloudigrade/cloudigrade/issues/355, when building the queries to gather the necessary data, I realized that we really should have established a relationship between from the `InstanceEvent` to the `MachineImage` to indicate what image was seen at the time of the event. This PR does exactly that, and it migrates the old (and now redundant) `AwsInstanceEvent.ec2_ami_id` in the process.


infinitewarp:
@cloudigrade/quality-engineering This change removes `ec2_ami_id` from the API that lists the instance events, and it replaces with a link to the image. If you guys have any tests that assert on that field specifically, we will need to update those...

cloudigrade/account/util.py
```diff
@@ -71,31 +74,48 @@ def save_instance_events(account, instance_data, region, events=None):
...
                 occurred_at=event['occurred_at'],
                 subnet=event['subnet'],
-                ec2_ami_id=event['ec2_ami_id'],
                 instance_type=event['instance_type'],
             )
             event.save()
     return instance
 
 
+# def get_or_create_awsmachineimage(account, ec2_ami_id):
```
abaiken: why is this commented out?

infinitewarp: Good catch. This is leftover from a previous experiment. Will delete in the next commit!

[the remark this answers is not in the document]
infinitewarp: Good catch. This is leftover from a previous experiment. Will delete in the next commit!

PR #407, opened 2018-07-09
Add check for empty messages in delete_message_from_queue

kholdaway:


cloudigrade/util/aws/sqs.py
```diff
@@ -39,6 +39,9 @@ def delete_message_from_queue(queue_url, messages):
         dict: The response from the delete call.
 
     """
+    if not bool(messages):
```
infinitewarp: Why are you explicitly casting this to a `bool`? `if not messages` should work for an empty list.

elyezer:
> Should we add a test?

:heart:

cloudigrade/util/aws/sqs.py
elyezer: I got curious about this too. But asked on Slack. Better to talk here instead.

ilyavw: In addition to adding a test for this, also adding an info/debug log would be helpful.

ilyavw:
https://github.com/cloudigrade/cloudigrade/pull/407#issuecomment-403512811
Shouldn't it be QE asking this question? 😉

cloudigrade/util/aws/sqs.py
kholdaway: I can remove that.  Great catch.

kholdaway: This happens every time the analyzer runs if instances haven't changed.  I can add this but it will be common.

ilyavw: Sounds like a debug log would be appropriate then.

kholdaway:
@abaiken @infinitewarp @ilyavw Back at you

cloudigrade/util/aws/sqs.py
```diff
@@ -39,6 +43,11 @@ def delete_message_from_queue(queue_url, messages):
         dict: The response from the delete call.
 
     """
+    if not messages:
+        logger.debug(
+            _('No messages to delete because messages list was empty.'))
```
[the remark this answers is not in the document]
kholdaway: @infinitewarp done.

elyezer:
https://github.com/cloudigrade/cloudigrade/pull/407#issuecomment-403512811
Shouldn't it be QE asking this question? 😉

We were checking if Devs would ask for it :1st_place_medal:

PR #414, opened 2018-07-10
Analyze logs for added/removed AMI tags

kholdaway:


cloudigrade/analyzer/tasks.py
```diff
@@ -118,7 +132,58 @@ def _parse_log_for_ec2_events(log):
...
+                        ami.tags.add(openshift_tag)
+                    else:
+                        logger.info(_(
+                            'Removing openshift tag from AMI {}').format(
+                                ami_id))
+                        ami.tags.remove(openshift_tag)
+                    ami.save()
+                else:
+                    logger.info(
+                        'Tag create/delete event referenced AMI {}, '
```
ilyavw: Looks like you forgot to throw a `.format()` in here. Also, your other logs messages appear to be run through gettext, but not this one?

cloudigrade/analyzer/tasks.py
```diff
@@ -118,7 +132,58 @@ def _parse_log_for_ec2_events(log):
...
+                        'but no AMI with this ID is known to cloudigrade.')
+
+
+def _is_valid_event(record, valid_events):
```
ilyavw: Was there a reason for this name change? The old name was pretty descriptive, but the new one is a little confusing as it doesn't seem to necessarily be validating if its valid, since it only checks a small subset (on/off events), not all events.

kholdaway: Yes. I used it to filter the tag create/delete events too.

kholdaway: Yes. I used it to filter the tag create/delete events too.

kholdaway: Given it took a list of valid events and I needed the exact same functionality, it made sense to rename it and reuse it for my events too.

cloudigrade/analyzer/tasks.py
kholdaway: @ilyavw Thanks!  Fixed..

kholdaway:
Demo video --> https://drive.google.com/open?id=11PYVUCnWzJRRgvUFel83X2bI-LIMrYLG

ilyavw:
Closing due to move to gitlab.

Issue #365, opened 2018-06-21
Marketplace image copying fails


infinitewarp:
## Summary

If a user is running an instance from a public marketplace image, our inspection of that image will fail unexpectedly because you are not allowed to copy marketplace images. It fails in a way that leaves our system thinking that the image has not yet been inspected (it's stuck in the "preparing" status), and that means we will be reporting it incorrectly and we will attempt inspection again in the future and continue failing.

## Steps to Reproduce

1. In a customer AWS account, launch an EC2 instance using the "Red Hat Enterprise Linux" on the quick start page.
1. Wait for the instance to finish starting completely.
1. Register the account with cloudigrade.
1. Start watching the Celery logs after registration successfully completes.

## Expected Result

- Somehow we recognize the image as belonging to the marketplace, we skip over it, and we set the image state in our database to "inspected" so we don't try to inspect it in the future.

## Actual Result

- Celery task fails with errors like this:
```
[2018-06-28 22:41:24,871: ERROR/ForkPoolWorker-3] Task account.tasks.copy_ami_snapshot[34253257-1b03-452b-b10b-a5f3f2642e72] raised unexpected: RuntimeError("Unexpected AWS error <class 'botocore.exceptions.ClientError'>: An error occurred (InvalidSnapshot.NotFound) when calling the DescribeSnapshots operation: The snapshot 'snap-02196f4f8507c9598' does not exist.",)
Traceback (most recent call last):
  File "/opt/cloudigrade/util/aws/helper.py", line 189, in wrapped
    result = original_function(*args, **kwargs)
  File "/opt/cloudigrade/account/tasks.py", line 57, in copy_ami_snapshot
    if customer_snapshot.encrypted:
  File "/opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/boto3/resources/factory.py", line 339, in property_loader
    self.load()
  File "/opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/boto3/resources/factory.py", line 505, in do_action
    response = action(self, *args, **kwargs)
  File "/opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/boto3/resources/action.py", line 83, in __call__
    response = getattr(parent.meta.client, operation_name)(**params)
  File "/opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/botocore/client.py", line 314, in _api_call
    return self._make_api_call(operation_name, kwargs)
[...]

PR #415, opened 2018-07-10
Handle marketplace images

ilyavw:
#365 

This adds logic around detecting marketplace images and marking them as inspected.


ilyavw:
Closing due to move to gitlab.

PR #793, opened 2021-02-05
Fix bug with account update

werwty:
If the ARN being updated has a different aws account id we delete the
clount and recreate it. With the recent change to delete the User if
the user doesn't have any more clounts, the update fails with a
Username not found message to sources.

This commit fixes this issue

### DIFF
--- a/cloudigrade/api/clouds/aws/util.py
+++ b/cloudigrade/api/clouds/aws/util.py
@@ -5,6 +5,7 @@ from decimal import Decimal
 
 from botocore.exceptions import ClientError
 from django.conf import settings
+from django.contrib.auth.models import User
 from django.db import IntegrityError, transaction
 from django.utils.translation import gettext as _
 from rest_framework.serializers import ValidationError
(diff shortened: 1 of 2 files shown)


cloudigrade/api/clouds/aws/util.py
```diff
@@ -1051,6 +1052,14 @@ def update_aws_cloud_account(
         )
         cloud_account.delete()
 
+        with transaction.atomic():
+            user, created = User.objects.get_or_create(username=account_number)
```
infinitewarp, line 1056: Just a general observation "thinking out loud" here. I think this is the the third instance in our code where we are `get_or_create`ing a user like this, and this is starting to feel like something we should wrap up in a helper function for more consistent reuse, testing, mocking, etc. I'm not suggesting you go change this now. Just planting the suggestion for maybe later. :)

infinitewarp (approved):

cloudigrade/api/clouds/aws/util.py
werwty: :+1: That sounds good, I'll merge this PR as is for now, because I've been waiting forever to close this story. But I'll add the refactoring to my todo list :)

PR #794, opened 2021-02-05
poetry-core 1.0.1 is currently broken

werwty:
Let's pin it to 1.0.0 for now, and revert this commit once it is fixed.

See also: https://github.com/python-poetry/poetry/issues/3650


katherine-black (changes requested):
Is there a source or any kind of information you can find behind this work around? Maybe the poetry issue link?

infinitewarp (approved):

katherine-black (approved):
I swear it wasn't there at first, :rubber-stamp:

Issue #738, opened 2021-01-29
Ensure AWS environment is healthy upon startup


katherine-black:
In GitLab by @infinitewarp on Sep 28, 2020, 15:31

## Important Note:

We are holding this issue from completing its implementation until we have either a) strong guidance for a solution regarding shared secrets/configs from the platform or b) March 26th has arrived which means we've given several weeks opportunity for conversation and feedback. If we have no new or strong guidance by March 26, let's proceed with whatever "ugly" but working solution we can identify (such as shoving configs in an s3 bucket).

## Summary

As a cloudigrade sysop, I want an automated process to run upon startup (alongside DB migrations) that ensures my AWS environment is appropriately configured and available so that cloudigrade can more seamlessly handle AWS environment changes. Currently, we have manual steps for setting up things in AWS that include defining objects and policies in ECS, SQS, and S3. We should automate their setup and verification upon cloudigrade deployment.

Aside from generally making our lives easier, this will become necessary as we move towards ephemeral environments and if/when the platform requests changing our main AWS account. The latter has been proposed at least once to me as a contingency for occasional but unexpected AWS service outages.

## Acceptance Criteria

- [x] Verify that cloudigrade automatically configures an "empty" AWS account with:
  - [ ] ECS cluster (including Auto Scaling Group and Launch Configurations) for houndigrade
  - [ ] Security Group for the Launch Configuration
  - [ ] SQS queue and S3 bucket with appropriate policies for CloudTrail
  - [ ] ...and anything else relevant we think of during development.
- [x] Verify that cloudigrade automatically REconfigures a "not-empty" AWS account with the same as above
- [x] Verify that cloudigrade alerts Slack after the deployment process finishes, with a brief status of the run
- [ ] QE to rerun IQE regression tests.

## Assumptions and Questions

- Probably no specific QE efforts on this; no automation tests.
- Should we "start small" here by simply migrating and modernizing our ooooooold Ansible scripts that are used for our review environments?
  - Yes, try this!
- If we see that an account appears to be set up but is set up **incorrectly**, should we attempt to repair? Or should we stop and alert as a failure/error?
  - No, we blindly rerun Ansible scripts. See next point.
- What is the order of operations here?
  - 1. We should run Ansible scripts in dry-run/verify mode FIRST.
  - 2. Always alert Sentry (consider adding a new Slack channel) with the results. Future iteration might make this conditional.
[...]

PR #802, opened 2021-02-15
Ensure AWS env is healthy part 1: Deployment changes

katherine-black:
#738


.github/workflows/push.yml
```diff
@@ -0,0 +1,170 @@
...
+      - uses: actions/checkout@v2
+      - run: apk --no-cache add git
+      - run: mkdir -p /opt && cd /opt
+      - run: wget https://github.com/openshift/source-to-image/releases/download/${S2I_VERSION}/${S2I_RELEASE}.tar.gz -O /opt/s2i.tar.gz
+      - run: tar --strip-components=1 -xzvf /opt/s2i.tar.gz -C /opt/
+      - run: mv /opt/s2i /usr/bin/
+      - env:
+          REDHAT_REGISTRY_LOGIN: ${{ secrets.REDHAT_REGISTRY_LOGIN }}
+        run: ${REDHAT_REGISTRY_LOGIN}
+      - run: s2i build --context-dir=${CONTEXT_DIR} --ref ${GITHUB_REF##*/} ${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY} ${NGINX_S2I_IMAGE} ${NGINX_IMAGE_NAME}:latest
```
infinitewarp, line 81: In `${GITHUB_REF##*/}`, what does the `##*/` mean or do or translate to? I'm not familiar with this style of syntax.

werwty (approved):

.github/workflows/push.yml
katherine-black: I'm doing some weird parameter expansion here, effectively I just want everything after the last slash, as that value is normally fed by github as `ref/heads/potatoes` or something along those lines.

Here is some more info: https://stackoverflow.com/questions/19482123/extract-part-of-a-string-using-bash-cut-split/19482947#19482947

infinitewarp (approved):

PR #814, opened 2021-02-24
738 ensure aws

katherine-black:


pyproject.toml
```diff
@@ -7,6 +7,8 @@ license = "GPL-3.0-or-later"
 
 [tool.poetry.dependencies]
 python = "^3.8"
+ansible = "~=2.10.5"
+boto = "~=2.49.0"
```
infinitewarp, line 11: Is this old `boto` version required somewhere by the Ansible stuff?

katherine-black: Yes 😬

deployment/playbooks/roles/aws/templates/env.j2
```diff
@@ -0,0 +1,17 @@
...
+
+AWS_ACCESS_KEY_ID="{{iam_user.user_meta.access_keys[0].access_key_id}}"
+AWS_SECRET_ACCESS_KEY="{{iam_user.user_meta.access_keys[0].secret_access_key}}"
+
+AWS_SQS_ACCESS_KEY_ID="{{iam_user.user_meta.access_keys[0].access_key_id}}"
+AWS_SQS_SECRET_ACCESS_KEY="{{iam_user.user_meta.access_keys[0].secret_access_key}}"
+
+AWS_NAME_PREFIX="{{env}}-"
+
+AWS_CLOUDTRAIL_EVENT_URL="{{ sqs_queue.queue_arn | regex_replace('^arn.+sqs.+us-east-1:(?P<account_id>\d+):(?P<sqs_queue_name>\S+)$', 'https://sqs.us-east-1.amazonaws.com/\\g<account_id>/\\g<sqs_queue_name>')}}"
```
infinitewarp, line 14: This looks like it works fine, but I'm trying to think if there's a more elegant way to set this value.

Do you think there's something we can do with [amazon.aws.aws_caller_info](https://docs.ansible.com/ansible/latest/collections/amazon/aws/aws_caller_info_module.html#ansible-collections-amazon-aws-aws-caller-info-module) to get the AWS account ID without having to parse apart an ARN like this?

infinitewarp: Also I'm wondering if we should make the region configurable, though I know we also have `us-east-1` and `us-east-1a` baked in several other places.

cloudigrade/config/settings/prod.py
```diff
@@ -2,11 +2,24 @@
...
+
+CLOUDIGRADE_ENVIRONMENT = env("CLOUDIGRADE_ENVIRONMENT")
+
+HOUNDIGRADE_AWS_AVAILABILITY_ZONE = "us-east-1a"
+HOUNDIGRADE_AWS_AUTOSCALING_GROUP_NAME = f"cloudigrade-ecs-asg-{CLOUDIGRADE_ENVIRONMENT}"
+HOUNDIGRADE_ECS_CLUSTER_NAME = f"cloudigrade-ecs-{CLOUDIGRADE_ENVIRONMENT}"
+
+AWS_NAME_PREFIX = f"{CLOUDIGRADE_ENVIRONMENT}-"
+
+AWS_CLOUDTRAIL_EVENT_URL = f"https://sqs.us-east-1.amazonaws.com/{_get_primary_account_id()}/cloudigrade-cloudtrail-s3-{CLOUDIGRADE_ENVIRONMENT}"
```
infinitewarp, line 20: Calling our code _during_ config loading feels fragile and risky to me. I checked the sequence of imports, and it _should_ be okay for now, but if we aren't very careful around the `sts` and `arn` packages, we could unknowingly introduce something like an import loop that we wouldn't see until the new code fails to start in "prod" (yes this config is also used for stage, etc). If we introduce something later in the chain needs to use something from `from django.conf import settings`, I suspect things could go badly here.


I pondered at https://github.com/cloudigrade/cloudigrade/pull/814/files#r614997276 if we could ask `aws_caller_info` for the AWS account ID. Should we consider doing something like that and just explicitly set an env var with the AWS account ID? On one hand, that would generally be redundant with the keys. On the other hand, "explicit is better than implicit"?

Any other thoughts or possible alternative solutions?

deployment/playbooks/roles/aws/templates/env.j2
```diff
@@ -0,0 +1,17 @@
+DJANGO_DEBUG="False"
+HOUNDIGRADE_AWS_AVAILABILITY_ZONE="{{ec2_asg_az}}"
+HOUNDIGRADE_AWS_AUTOSCALING_GROUP_NAME="{{ec2_autoscaling_group.auto_scaling_group_name}}"
+HOUNDIGRADE_ECS_CLUSTER_NAME="{{ecs_cluster.cluster.clusterName}}"
```
infinitewarp, line 4: Should this and the lines above it be deleted since we no longer use these env vars in the app?

Actually, now that I think about it, is this env file just an artifact from when we were hoping to put in the PVC?

deployment/playbooks/roles/aws/tasks/iam.yml
```diff
@@ -0,0 +1,58 @@
...
+      iam_type: user
+      iam_name: "{{ iam_user_name }}"
+      policy_name: "{{ iam_policy_name }}"
+      policy_json: "{{ lookup( 'template', 'cloudigrade-deployment-policy.json.j2') | to_json }}"
+      state: present
+
+  - name: create / write deploy env vars
+    template:
+      src: env.j2
+      dest: /mnt/secret_store/.env
```
infinitewarp, line 43: How is this used if we've not going to use the persistent volume?

deployment/playbooks/roles/aws/templates/env.j2
katherine-black: Yes, that is the env file template.

deployment/playbooks/roles/aws/templates/env.j2
katherine-black: There is no elegant way to set this value because the aws module fails to return the url like it is supposed to half the time. I don't really see the value in having an extra call to fetch information that we can otherwise put together with what was already returned.

Additionally, this file is unused until we have shared storage between pods.

cloudigrade/config/settings/prod.py
katherine-black: Well we can't use Ansible to do it, because there is no way to get data from ansible to the deployment. Only vars we explicitly bake in, or set in the config map. "Setting" an account number to a set value once again drives us further away from being able to just get aws keys and run in any environment.

I dislike this strongly, but I see no other way to continue as I can't just provide a templated .env file that would be ready by all the pods.

deployment/playbooks/roles/aws/tasks/iam.yml
katherine-black: Its not, you'll notice that the iam.yml playbook is not included in the aws role.

katherine-black: https://github.com/cloudigrade/cloudigrade/pull/814/files#diff-30c24893354ebceec6700ad16bfd9342c72d221b3856e21c2c81f3342bed3020R6

deployment/playbooks/roles/aws/templates/env.j2
katherine-black: `.j2` == jinja

deployment/playbooks/roles/aws/tasks/iam.yml
infinitewarp: Then I think we should comment this out or at least add a clear inline comment explaining for future readers that it's not actually used and is a placeholder for future work. Otherwise I foresee it confusing anyone (not named Kat 😅) trying to step through understanding our configs.

deployment/playbooks/roles/aws/templates/env.j2
infinitewarp: Even if we keep the template in hopes that "future us" will find a way to make it relevant, these actual values (`DJANGO_DEBUG`, `HOUNDIGRADE_AWS_AVAILABILITY_ZONE`, `HOUNDIGRADE_AWS_AUTOSCALING_GROUP_NAME`, and `HOUNDIGRADE_ECS_CLUSTER_NAME`) appear not to be used for deployments, and if that's correct, I think they should be culled from the templated env file. We do load these value names in `base.py` (like [here](https://github.com/cloudigrade/cloudigrade/blob/959d956f4440513542c37d581fd7cb0a1ecce1dc/cloudigrade/config/settings/base.py#L124)), but then we immediately overwrite them with values in `prod.py` [here](https://github.com/cloudigrade/cloudigrade/pull/814/files#diff-77fcf72288500c6f144a8a8c99afd446dde68815c69f27edd3d7358389a80d1bR14).

I assume this generated env file would only be used in "real" deployments, not locally, which would mean these values will _always_ get overwritten by their `prod.py` counterparts. Or do you envision a way that this file could be generated and used for local development where `local.py` _doesn't_ overwrite the values from the env vars/file? Or have I misread the intention of this file?

katherine-black:
@infinitewarp I've updated the branch and removed all the confusing left behind code.

infinitewarp (dismissed):

katherine-black:
@cloudigrade/engineering this should be the final update barring any feedback changes. The secret for the slack webhook is already set in all environments.

infinitewarp (approved):
What could possibly go wrong? 🙂

🚢 🇮🇹

Issue #737, opened 2021-01-29
Use new AWS API to load EC2 instance type definitions


katherine-black:
In GitLab by @infinitewarp on Sep 28, 2020, 12:04

## Summary

As a cloudigrade developer and sysop, I want my EC2 instance mapping to reload quickly and reliably. Currently, we use the pricing API to get the EC2 instance types, but this is very slow and fetches lots of unnecessary and redundant data for our needs. AWS recently added a new API *specifically* to get EC2 instance type descriptions: `describe-instance-types`. We should use that.

Example from the `aws` CLI:

```
$  aws ec2 describe-instance-types | head -n 30
{
    "InstanceTypes": [
        {
            "InstanceType": "m5ad.large",
            "CurrentGeneration": true,
            "FreeTierEligible": false,
            "SupportedUsageClasses": [
                "on-demand",
                "spot"
            ],
            "SupportedRootDeviceTypes": [
                "ebs"
            ],
            "SupportedVirtualizationTypes": [
                "hvm"
            ],
            "BareMetal": false,
            "Hypervisor": "nitro",
            "ProcessorInfo": {
                "SupportedArchitectures": [
                    "x86_64"
                ],
                "SustainedClockSpeedInGhz": 2.2
            },
            "VCpuInfo": {
                "DefaultVCpus": 2,
                "DefaultCores": 1,
                "DefaultThreadsPerCore": 2,
                "ValidCores": [
                    1
```

Results from the API are limited by region. So, we'll still need to iterate the request over each of our supported regions. The corresponding boto3 code would look something like:

```
import boto3
region_name = 'us-east-1'
ec2_client = boto3.client('ec2', region_name=region_name)
results = ec2_client.describe_instance_types()
```

An example result looks like:

```
>>> results['InstanceTypes'][0]['InstanceType']
'i3en.12xlarge'
>>> results['InstanceTypes'][0]['VCpuInfo']
{'DefaultVCpus': 48, 'DefaultCores': 24, 'DefaultThreadsPerCore': 2, 'ValidCores': [2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22, 24], 'ValidThreadsPerCore': [1, 2]}
>>> results['InstanceTypes'][0]['MemoryInfo']
{'SizeInMiB': 393216}
```

## Acceptance Criteria

- [x] Verify the `loadec2definitions` management command still functions normally
[...]

PR #820, opened 2021-03-01
Use new AWS api to get instance definitions 

katherine-black:
https://github.com/cloudigrade/cloudigrade/issues/737


cloudigrade/api/migrations/0038_instancedefinition.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("api", "0037_delete_instance_definition"),
+    ]
+
+    operations = [
+        migrations.CreateModel(
```
infinitewarp, line 13: Since the table will be empty immediately following this deployment, if there are any concurrent usage calculation jobs that run shortly after this deployment, they will likely produce incorrect totals.

I think we need to make AWS calls to rebuild the definitions here in this migration or run `loadec2definitions` after `migrate` in the deployment mid-hook. Calling within the migration could be tricky since the model classes are in a weird state and you can't easily import and call our functions. The latter feels bad to introduce AWS as a dependency during all deployments, but it's much easier to implement and we could pull it back out after one deploy. Thoughts or other ideas?

katherine-black: [Thread](https://ansible.slack.com/archives/C8VGAPJNN/p1614712944056100?thread_ts=1614710040.053600&cid=C8VGAPJNN)

infinitewarp (approved):

Issue #719, opened 2021-01-29
HTTP delete accounts asynchronously


katherine-black:
In GitLab by @infinitewarp on Jul 22, 2020, 15:08

## Summary

As a cloudigrade client or sysop, I want HTTP requests to respond as fast as possible. Currently, HTTP deleting a CloudAccount may take many seconds if there are many related objects that also needs to be found and deleted. Consider instances, images, events, runs, and concurrent usage objects.

Let us simply wrap the `delete` call in an async celery function and call that function `.delay` and let it take care of itself. We don't currently need to worry about the customer hitting the API quickly while rows are being deleted and the customer seeing "stale" data; let us see if that's actually a problem later. If that's a problem, we can iterate and introduce some additional state (like "marked for deletion") to the account at a later time.

## Acceptance Criteria

- [x] Verify that HTTP deleting an account still works.
- [x] Verify that deleting an account via sources-api object deletion still works.
- [x] Verify that the deletion of an account is handled by a Celery worker asynchronously.
- [ ] QA verify using automation that the sources-api workflow continues to work.

## Assumptions and Questions

- Probably no new integration tests for QE. Probably just rerunning tests to check for regressions.
- This should be considered to have a relatively low priority because we do not actively promote the use of the HTTP API for creating and deleting accounts.
- However, this would still be useful even in the case where sources-api causes deletion. Would we want to very quickly mark the account to have its results filtered away so that API calls made between sources-api deletion and *actual* deletion no longer include the data? Maybe this is overkill. Maybe we don't care because nobody really uses the HTTP APIs, and the concurrent API is known accepted to be "delayed" (effectively) in its representation.
  - Don't worry about new state/status/etc. on the account for now.
- This is probably a very low priority and should only be done if we're looking for things to do or we learn something new and relevant about how our APIs are being used.

katherine-black:
In GitLab by @infinitewarp on Jul 23, 2020, 10:54

changed title from **HTTP delete accounts synchronously** to **HTTP delete accounts {+a+}synchronously**

infinitewarp:
See also: https://github.com/cloudigrade/cloudigrade/issues/828

We should try to ensure that we aren't letting accounts get orphaned first before we focus on the cleanup effort here.

infinitewarp:
- Demo of updated internal API flow: https://asciinema.org/a/400036
- Demo of existing sources-api Kafka flow: https://asciinema.org/a/400044

katherine-black:
[...]

PR #830, opened 2021-03-08
719 async delete

infinitewarp:
for https://github.com/cloudigrade/cloudigrade/issues/719


werwty (dismissed):
This looks good!

This is super super nitpicky, so please feel free to disregard- but I think the right status code to return now is a 202 (accepted), since the endpoint triggers an async job, and we don't know for sure if that job will complete successfully.

infinitewarp:
Excellent point @werwty! I hadn't thought of that, but you're right: `202` makes perfect sense here. I'll make this delete response use that instead of the default `204`.

katherine-black (approved):

Issue #768, opened 2021-01-29
ConcurrentUsage does not need to queue calculation tasks if given time is before user creation.


katherine-black:
In GitLab by @werwty on Dec 10, 2020, 10:11

## Summary

As a developer/QE tester I want cloudmeter's concurrentusage API to return fast for answers that are obvious. One case is if the user requests a concurrentusage date before their user was created. Instead of queueing up a calculation task, returning 425, and waiting 30 minutes to figure out that we have no data, we should shortcut this and just tell the user we have no data. 

## Acceptance Criteria

- [ ] Verify that a concurrentusage request with an end_date prior to user creation date returns an error response.

## Assumptions and Questions

- Example use cases:
  - User was created on 2020-12-10. Request has start and end dates both before 2020-12-10. Response should be an error stating that the end_date can not be prior to user creation date.
  - User was created on 2020-12-10. Request has start date before 2020-12-10 and end date after 2020-12-10. Response should follow the normal process which may include 425, task, etc. if not already calculated.
- Warning for automation tests: this effectively requires a new user, but that may be difficult given how users are tied to Red Hat accounts that we cannot freely create. However, @werwty is working on a related change now to automatically delete users when we no longer need them.
  - This means setup for this specific IQE test would need to clear out all clounts for a user (so the user is automatically deleted) *or* run _before_ any other tests would create a user *or* need to find a way to create whole new users.
  - Alternate solution if we have not deleted the user: when we release the new internal APIs, the test could ask the `users` API for its own user, get the created date from that, and set the dates for the test to before the created date. However, the qa environment does not (yet??) appear to have Turnpike running; so, we might not actually be able to use the internal APIs. :frowning2:

AC Updated after conversation w/ @werwty and @infinitewarp

infinitewarp:
QE issue tracked here: https://projects.engineering.redhat.com/browse/SMQE-294

katherine-black:
```
❯ http --verify=false --auth ${AUTH} --proxy="https:http://squid.corp.redhat.com:3128" https://cloud.stage.redhat.com/api/cloudigrade/v2/concurrent/ start_date==2017-01-01 end_date==2017-01-02

HTTP/1.1 400 Bad Request
Allow: GET, HEAD, OPTIONS
Connection: keep-alive
Content-Length: 59
Content-Type: application/json
Date: Thu, 11 Mar 2021 21:28:16 GMT
Server: openresty/1.13.6.1
[...]

PR #832, opened 2021-03-11
Return an error if end_date is before user create date.

katherine-black:
https://github.com/cloudigrade/cloudigrade/issues/768

### DIFF
--- a/cloudigrade/api/views.py
+++ b/cloudigrade/api/views.py
@@ -91,8 +91,8 @@ class DailyConcurrentUsageViewSet(viewsets.GenericViewSet, mixins.ListModelMixin
 
     def get_queryset(self):  # noqa: C901
         """Get the queryset of dates filtered to the appropriate inputs."""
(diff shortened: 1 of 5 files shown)


cloudigrade/api/views.py
```diff
@@ -91,8 +92,8 @@ class DailyConcurrentUsageViewSet(viewsets.GenericViewSet, mixins.ListModelMixin
 
     def get_queryset(self):  # noqa: C901
         """Get the queryset of dates filtered to the appropriate inputs."""
+        user = User.objects.get(id=self.request.user.id)
```
infinitewarp, line 95: Is `self.request.user` already a `User` instance or is it some other object type? Could you just reference it directly in the lines below like `self.request.user.date_joined.date()`?

katherine-black: Oh duh, good catch!

infinitewarp (approved):

PR #870, opened 2021-04-09
add psql cli tools to cloudigrade image

infinitewarp:
I'm adding these to help diagnose some issues in our deployed
environments. This might be temporary, but it might also be useful
to keep around for future investigations.

Yes, these curls and rpm commands are kind of ugly, but neither UBI
nor EPEL include appropriate postgresql packages. Also, we have to
tell rpm "--nodeps" because we don't want it to pull in systemd and
all of *its* dependencies (which postgresql11 declares but really
only needs if you want to run the psql *server*).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,14 @@ RUN microdnf install shadow-utils \
 	&& useradd -r cloudigrade \
 	&& microdnf clean all
 
+RUN microdnf install libicu \
+    && curl -O https://download.postgresql.org/pub/repos/yum/11/redhat/rhel-8.3-x86_64/postgresql11-libs-11.9-4PGDG.rhel8.x86_64.rpm \
+    && curl -O https://download.postgresql.org/pub/repos/yum/11/redhat/rhel-8.3-x86_64/postgresql11-11.9-4PGDG.rhel8.x86_64.rpm \
+    && rpm -iv postgresql11-libs-11.9-4PGDG.rhel8.x86_64.rpm \
+    && rpm --nodeps -iv postgresql11-11.9-4PGDG.rhel8.x86_64.rpm \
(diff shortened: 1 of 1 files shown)


abellotti:
I think this is fine, I built a minimal ubi8.3 with this diff with a dumb-init pod, was getting:

```
psql: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```

So it was installed 👍   That error could just be my pod.

abellotti (approved):

infinitewarp:
@abellotti Yeah, what you saw there is normal output for the bare `psql` command. If you don't give explicit connection parameters, it assumes you are trying to connect to the default `postgres`-name instance running locally and that you're connecting over the local socket instead of TCP. When I was going through this setup locally, I tested it by connecting from the `psql` command inside the container to a postgres server running outside the container on my Mac. It looked like:

```
[root@3797db6388ba cloudigrade]# psql -U postgres -h 192.168.1.137 postgres
psql (11.9, server 13.2)
WARNING: psql major version 11, server major version 13.
         Some psql features might not work.
Type "help" for help.

postgres=# \d
                                 List of relations
 Schema |                        Name                         |   Type   |  Owner
--------+-----------------------------------------------------+----------+----------
 public | api_awscloudaccount                                 | table    | postgres
 public | api_awscloudaccount_id_seq                          | sequence | postgres
 public | api_awsinstance                                     | table    | postgres
...
```

In this example output, by the way, the version mismatch mentioned here is appropriate. I install postgres 11 in the container because that's the version we use "for real" in the stage and prod RDS instances, but I just happened to be running a postgres 13 server on my Mac.

PR #887, opened 2021-04-29
update dependencies

infinitewarp:
We're pretty far our of date on a lot of our Python dependencies. This PR updates most of these (but not Ansible) to their latest stable versions. A few small updates to our code are also included here to deal with breaking changes. I skimmed all the relevant release notes I could find, and I _think_ we should be safe to proceed from here. However, before we merge this, I still need to do some local testing with a real database that was seeded by a version of cloudigrade with the _old_ dependencies.

I specifically excluded Ansible because verifying the playbook is a different enough task that I'd like to open a separate isolated PR for that later.


poetry.lock
```diff
@@ -1,13 +1,13 @@
 [[package]]
 name = "amqp"
-version = "2.6.1"
+version = "5.0.6"
```
abellotti, line 3: 👍  that puppy was really old.

katherine-black (dismissed):

abellotti (approved):

infinitewarp:
I've been manually testing this change locally today by doing (roughly) the following:

- `drop schema public cascade; create schema public;` to ensure my DB is clean 
- `git checkout master`, `poetry install`, and `manage.py migrate` to ensure I have the old code, dependencies, and models
- `manage.py createsuperuser` and `manage.py spawndata` to generate a bunch of data
- `manage.py runserver`, `celery worker`, make API requests, and watch the logs to ensure normal operation.
- `git checkout dependencies`, `poetry install`, and `manage.py migrate` to upgrade to new code and deps
- `manage.py runserver`, `celery worker`, make API requests, and watch the logs to ensure normal operation.

In general, things look good! I encountered only one problem:

```
You are using `--app` as an option of the worker sub-command:
celery worker --app celeryapp <...>

The support for this usage was removed in Celery 5.0. Instead you should use `--app` as a global option:
celery --app celeryapp worker <...>
```

So, we'll need an e2e-deploy update to rearrange the args in our `beat` and `worker` commands. Fortunately, it looks like the old version of celery already supports this new argument order requirement.

katherine-black (approved):

Issue #718, opened 2021-01-29
Use Kafka instead of HTTP to update sources-api status


katherine-black:
In GitLab by @infinitewarp on Jul 16, 2020, 14:44

## Summary

As a sysop, I want to reduce synchronous waiting operations in cloudigrade. sources-api recently introduced the ability to read from a Kafka topic (we write; they read) for updating an application availability status. We should use this Kafka topic instead of the HTTP API, and we should trust that once we've written to the topic, our end of the update contract has been fulfilled.

See also:

- https://projects.engineering.redhat.com/browse/TPINVTRY-970
- https://github.com/RedHatInsights/sources-api/pull/239

## Acceptance Criteria

- [ ] Verify through code review that cloudigrade no longer uses the sources-api HTTP API to set the availability  status on applications.
- [ ] Verify that creating a cloud account (through sources-api) with valid AWS setup results in sources-api being updated with "available" status. This is the happy path regression test!
- [ ] Verify that creating a cloud account (through sources-api) with invalid AWS setup results it results in sources-api being updated with "unavailable" status. This is the unhappy path regression test!
  - Here are some ideas for making an invalid AWS setup at the time of sources-api object creation:
    - a well-formed ARN that has bogus values (e.g. `arn:aws:iam::000000000000:role/potato-role`)
    - a ARN for a Role that does not actually have a trusted relationship with our AWS account
    - a ARN for a Role that has a policy which is missing some of our required permissions that we verify during setup (e.g. `ec2:CopyImage` is missing)
  - Most if not all of those examples are already included in QE's automated regression test suite.
- [ ] Verify that the above criteria are correct via QE automation.

## Assumptions and Questions

- Implementation should be using an asynchronous celery task that retries using our typical exponential backoff. This should help us confidently deliver the messages even if Kafka is misbehaving and not accepting out messages.
- We do not current do "chaos testing" where we break Kafka. So, code review and unit testing should be sufficient for verifying the retry behavior of the aforementioned celery task.

abellotti:
Pull requests:
- [ ] https://github.com/cloudigrade/cloudigrade/pull/890
- [ ] https://github.com/RedHatInsights/e2e-deploy/pull/3051

abellotti:
was closed by the pr merge, re-opening until things are built/deployed/verified.

abellotti:
@infinitewarp well, seems to be processing them in CI and all happy paths, no KafkaException or BufferError.

[...]

PR #890, opened 2021-05-05
Use Kafka for sources application availability updates

abellotti:
Updated cloudigrade so that it uses Kafka for requesting application availability status updates instead of doing HTTP calls
to sources.  New kafka based notifications are celery backed and will retry upon failure to send the messages to Kafka.

For https://github.com/cloudigrade/cloudigrade/issues/718

- [x] a bit more sanity test
- [x] additional tests for codecov 
- [x] e2e deploy PR for the new kafka topic used - https://github.com/RedHatInsights/e2e-deploy/pull/3051


abellotti:
So far happy with the sanity tests

- With sources down, but Kafka up, did a notify task delay, things were queued in Kafka, then I manually started sources, he got the message and updated the application availability status.

- With sources up, but with zookeeper down, did a notify task delay, things failed, but were being retried by the celery worker. I manually restart zookeeper and kafka, message went through and sources got the message and updated the application availability status.

cloudigrade/api/error_codes.py
```diff
@@ -30,12 +29,13 @@ def get_message(self):
         """Get the external message for an error."""
         return self.message % {"error_code": self.code}
 
-    def notify(self, account_number, application_id, error_message=None):
+    def notify(self, application_id, error_message=None):
         """Tell sources an application is not available because of error."""
+        from cloudigrade.api.tasks import notify_application_availability_task
```
infinitewarp, line 34: the `cloudigrade.` root on this and similar import paths (like in `models.py` and the various tests) should not be necessary, and it may actually cause problems because the `PYTHONPATH` for our service should already be at that location.

abellotti: 👍  much nicer and shorter.  Thank goodness for PyCharm find and replace 😄

cloudigrade/util/redhatcloud/sources.py
```diff
@@ -217,46 +224,40 @@ def notify_application_availability(
...
+        kafka_producer = KafkaProducer(sources_kafka_config)
 
-    if response.status_code == http.HTTPStatus.NOT_FOUND:
-        logger.info(
-            _(
-                "Cannot update availability status, application id "
-                "%(application_id)s not found."
-            ),
-            {"application_id": application_id},
+        kafka_producer.poll(0)
```
infinitewarp, line 249: What is the purpose of this `poll(0)`?

cloudigrade/util/redhatcloud/sources.py
```diff
@@ -217,46 +224,40 @@ def notify_application_availability(
...
-    elif response.status_code != http.HTTPStatus.NO_CONTENT:
-        message = _(
-            "Unexpected status {status} updating application "
-            "{application_id} status at {url}"
-        ).format(status=response.status_code, application_id=application_id, url=url)
-        raise SourcesAPINotOkStatus(message)
+        kafka_producer.flush()
+    except BufferError as error:
+        message = f"BufferError: {str(error)}"
+        logger.info(message)
```
infinitewarp, line 258: Are the `BufferError` and `KafkaException` exceptions expected to be raised in unusual error cases? If so, I'd be tempted to change this `logger.info` and the one below to `logger.exception`. That would result in a more immediate Sentry alert and would include the stack trace in the message that might help diagnose where things went wrong in the guts of `kafka_producer`'s calls.

cloudigrade/util/tests/redhatcloud/test_sources.py
```diff
@@ -20,6 +25,11 @@ def setUp(self):
         self.account_number = str(_faker.pyint())
         self.authentication_id = _faker.user_name()
         self.application_id = _faker.pyint()
+        self.listener_server = "test_kafka_server"
+        self.listener_port = "19092"
+        self.sources_resource_type = "test_application"
+        self.sources_kafka_topic = "test_availability_status_topic"
+        self.sources_availability_event_type = "test_availability_status"
```
infinitewarp, line 32: Since this test suite already has a Faker instance being used just above this line, let me introduce you! 😀 

https://github.com/joke2k/faker#basic-usage

We try to avoid entering hard-coded values into our tests unless a specific situation calls for a persistent known value. This helps to ensure that we aren't _accidentally_ relying on the behavior of some static value in some unusual way. One possible use of Faker here might look like:

```
self.listener_server = _faker.hostname()
self.listener_port = str(_faker.pyint())
self.sources_resource_type = _faker.slug()
self.sources_kafka_topic = _faker.slug()
self.sources_availability_event_type = _faker.slug()
```

cloudigrade/util/tests/redhatcloud/test_sources.py
```diff
@@ -123,63 +133,151 @@ def test_get_sources_cloudigrade_application_type_fail(self, mock_get):
...
+        """Test notify source application availability handling KafkaException."""
+        application_id = _faker.pyint()
+        availability_status = "available"
 
-        mock_patch.assert_called()
+        sources_kafka_config = {
+            "bootstrap.servers": f"{self.listener_server}:{self.listener_port}"
+        }
+
+        kafka_producer = mock_kafka_producer(sources_kafka_config)
```
infinitewarp, line 253: I see these same lines being repeated in several tests. I think they might be a good candidate for moving into `setUp` and defined as instance variables on `self` to keep the individual tests DRY.

Ditto for the `payload = {` definition a few lines below.

infinitewarp (changes requested):
Sorry it took a while to get around to a full read of this. I added a few questions and comments along the way, but in general, I think this is on the right track.

cloudigrade/util/redhatcloud/sources.py
abellotti: No longer necessary, had it in the early prototype sample code.  Only useful if we register a callback. The flush we do a couple of lines later is all we need.  will remove.

cloudigrade/util/redhatcloud/sources.py
abellotti: logger.exception it is 👍

infinitewarp (dismissed):
👍

infinitewarp (approved):

infinitewarp (approved):

Issue #894, opened 2021-05-06
make "yesterday" default for concurrent API


infinitewarp:
## Summary

As the tally service, I want to request concurrent usage totals for a user only once per day and be confident that the calculated data is stable. Currently, the default behavior for the concurrent API is to return totals for "today", but since totals could change between the time of that call and 23:59:59.999, the new totals for that day may never be seen by tally.

Instead, we should return "yesterday" as the default when no argument are given, and any request that specifies "today" or later should be considered invalid. Currently, we allow "today" and only treat _future_ dates as invalid. By returning "yesterday" data, we have much higher confidence that our calculated totals will will have the complete activity for that date.

On 2021-05-06, @infinitewarp asked @kahowell in the Subscription Watch sync-up meeting if changing cloudigrade's default behavior to return "yesterday" would make sense, and the consensus was that *yes we should proceed with this change.* The tally service is aware of the optional date parameters for our concurrent API, but their client currently only uses `null` for each of them.

## Acceptance Criteria

- [ ] Verify concurrent API request with no arguments returns data for "yesterday"
  - For example, a request made at `2021-05-06 15:12:54` would return totals for the date `2021-05-05`.
- [ ] Verify concurrent API request with `start_date` set to "today" or later returns a 400 error response with an explanation.
  - We currently return an error message like "start_date cannot be in the future" for future dates. This message should update to indicate that the current date is also not allowed.
- [ ] Verify we have an internal API that generally behaves the same as the public "concurrent" API, but the internal version has no date restrictions.
  - This should enable QE to keep testing with "today" data.
- [ ] Verify that any related docs (such as REST API examples, openapi.json) are updated.

## Assumptions and Questions

- Kevin said that tally's current schedule is to ask for concurrent usage data at 01:00 UTC. This should generally allow sufficient time for slow AWS CloudTrail messages to be received and processed by cloudigrade, but there's also the possibility that at the end of the day we get a "power_on" event for an instance with a brand new image we've never seen and need to inspect, but the inspection process for that new instance+image could take an hour or longer. In this edge case, we still might be updating yesterday's totals _after_ tally has retrieved them. We should start thinking about how to address this problem in the future. Maybe tally should ask for the last two days and update their records? Maybe we should always return yesterday *and* the previous day?
[...]

PR #898, opened 2021-05-12
[WIP] Default to yesterday for the /concurrent/ API

abellotti:
- Updated the /concurrent/ API to return yesterday's daily concurrent usage instead of today, changing the default behavior of
  /api/cloudigrade/v2/concurrent/
- Requesting a start_date of today with the concurrent api now returns a 400.
- The current behavior of the /concurrent/ API is made available via the new /internal/api/cloudigrade/v1/concurrent endpoint.
- Updated tests for public concurrent API for the new defaults.
- Updates to generated docs and openapi

Solves #894 

Marking as WIP
- [ ] sanity check/review


cloudigrade/internal/views.py
```diff
@@ -508,3 +511,52 @@ class InternalAzureMachineImageViewSet(
         "created_at": ["lt", "exact", "gt"],
         "updated_at": ["lt", "exact", "gt"],
     }
+
+
+class InternalDailyConcurrentUsageViewSet(
```
infinitewarp, line 516: It looks like this class and `DailyConcurrentUsageViewSet` are copy-pasted from each other with only minor changes. We should consolidate them to stay DRY. Consider having one subclass the other and add helper methods for the relevant differences.

cloudigrade/internal/tests/views/test_dailyconcurrentusageviewset.py
```diff
@@ -0,0 +1,312 @@
...
+from django.utils.translation import gettext as _
+from rest_framework.test import APIClient, APIRequestFactory
+
+from api.models import ConcurrentUsageCalculationTask
+from api.tests import helper as api_helper
+from util.misc import get_today, get_yesterday
+from util.tests import helper as util_helper
+
+
+class InternalDailyConcurrentUsageViewSetTest(TransactionTestCase):
```
infinitewarp, line 16: Like with `DailyConcurrentUsageViewSet` and `InternalDailyConcurrentUsageViewSet`, this looks like a copy-paste job. If we change the the viewset classes so that one extends the other, we don't need to duplicate all these tests. We can simply have the tests for the subclass check only the things that are different in the subclass.

infinitewarp (changes requested):
See inline comments about duplicated code.

abellotti:
@brad duplicated code gone, shared common tests between the two and using the same SharedDailyConcurrentUsageViewSet just with different initialization.

cloudigrade/api/tests/views/test_dailyconcurrentusageviewset.py
```diff
@@ -1,229 +1,25 @@
...
-        client = APIClient()
-        client.force_authenticate(user=self.user1)
-        response = client.get(
-            "/api/cloudigrade/v2/concurrent/", data=data, format="json"
-        )
-        self.assertEqual(response.status_code, 400)
-
-        body = response.json()
-        self.assertEqual(body["end_date"], [_("end_date cannot be in the future.")])
+        SharedDailyConcurrentUsageViewSetTest.setUp(self)
```
infinitewarp, line 22: The proper convention in Python for calling the parent class is to use `super` like this instead of directly referencing it:
```suggestion
        super(DailyConcurrentUsageViewSetTest, self).setUp()
```

cloudigrade/api/tests/views/test_dailyconcurrentusageviewset.py
```diff
@@ -1,229 +1,25 @@
...
+    SharedDailyConcurrentUsageViewSetTest,
+)
 from util.misc import get_today
-from util.tests import helper as util_helper
 
 
-class DailyConcurrentUsageViewSetTest(TransactionTestCase):
+class DailyConcurrentUsageViewSetTest(
+    SharedDailyConcurrentUsageViewSetTest, TransactionTestCase
+):
```
infinitewarp, line 17: Since `SharedDailyConcurrentUsageViewSetTest` already extends `TransactionTestCase`, this shouldn't be necessary.

```suggestion
class DailyConcurrentUsageViewSetTest(SharedDailyConcurrentUsageViewSetTest):
```

cloudigrade/internal/tests/views/test_dailyconcurrentusageviewset.py
```diff
@@ -0,0 +1,65 @@
...
+from api.tests import helper as api_helper
+from api.tests.views.shared.test_dailyconcurrentusageviewset import (
+    SharedDailyConcurrentUsageViewSetTest,
+)
+from util.misc import get_today
+
+
+class InternalDailyConcurrentUsageViewSetTest(
+    SharedDailyConcurrentUsageViewSetTest, TransactionTestCase
+):
```
infinitewarp, line 17: Like with `DailyConcurrentUsageViewSetTest`, since `SharedDailyConcurrentUsageViewSetTest` already extends `TransactionTestCase`, this shouldn't be necessary.

```suggestion
class InternalDailyConcurrentUsageViewSetTest(SharedDailyConcurrentUsageViewSetTest):
```

cloudigrade/internal/tests/views/test_dailyconcurrentusageviewset.py
```diff
@@ -0,0 +1,65 @@
...
+    """InternalDailyConcurrentUsageViewSet test case."""
+
+    def setUp(self):
+        """Set up a bunch of test data."""
+        SharedDailyConcurrentUsageViewSetTest.setUp(
+            self, concurrent_api_url="/internal/api/cloudigrade/v1/concurrent/"
+        )
```
infinitewarp, line 24: Like with `DailyConcurrentUsageViewSetTest`,  using `super` here would look like:
```suggestion
        super(InternalDailyConcurrentUsageViewSet, self).setUp(concurrent_api_url="/internal/api/cloudigrade/v1/concurrent/")
```

cloudigrade/api/views.py
```diff
@@ -75,49 +75,71 @@ def list(self, *args, **kwargs):
...
     schema = schemas.ConcurrentSchema(tags=["api-v2"])
     serializer_class = serializers.DailyConcurrentUsageSerializer
 
-    def get_queryset(self):  # noqa: C901
+    def get_queryset(
+        self,
+        default_start_date=None,
+        default_end_date=None,
+        latest_start_date=None,
+        latest_end_date=None,
```
infinitewarp, line 89: I believe Django apps tend not to define and use custom arguments in functions like `get_queryset` because that could conflict with how the default view classes load them, which may not expose an interface to provide additional arguments. Although the implementation here probably works just fine in our case, I don't know if it looks like a Django "best practice".

Instead of overriding the call to `get_queryset` in the child classes, what would you think about the following ideas?

We could define separate helper methods (_maybe_ decorated as properties, but I don't have as strong feelings about that here) that `get_queryset` would use for each of the variables that differs between the child classes. Then the parent `get_queryset` here would call _those_ methods, and any child classes would _only_ implement those methods, not the `get_queryset` method itself. 

Also, we could drop the `SharedDailyConcurrentUsageViewSet` distinction entirely since it probably doesn't need to exist. I think if we were to diagram out the use and inheritance tree, `SharedDailyConcurrentUsageViewSet` looks and behaves more like an abstract base class since nothing instantiates it directly and only its child classes are instantiated. We could flatten the relationship a bit so that only `DailyConcurrentUsageViewSet` and `InternalDailyConcurrentUsageViewSet` exists, with the latter being a child of the former and the latter only implementing those extra getter/property methods.

Combining those ideas, the implementation could look like this pseudocode:

```
class DailyConcurrentUsageViewSet(viewsets.GenericViewSet, mixins.ListModelMixin):
  @property
  def latest_start_date(self):
    return get_today() - timedelta(days=1)

  def get_queryset(self):
    ...
    latest_start_date = self.latest_start_date
    ...

class InternalDailyConcurrentUsageViewSet(DailyConcurrentUsageViewSet):
  @property
  def latest_start_date(self):
    return get_today()
```

abellotti: yea, that was one of the possibilities (class methods instead of parameters), I'll give it a whirl. 👍

infinitewarp (approved):
Thank you for graciously handling all those review comments. Looks good! 👍

PR #902, opened 2021-05-14
Cache value of get_sources_cloudigrade_application_type_id

### LINKED ISSUE #663: Cache/memoize `get_sources_cloudigrade_application_type_id`
In GitLab by @infinitewarp on Mar 24, 2020, 18:06

## Summary

As a cloudigrade sysop, I want to reduce unnecessary network calls so that cloudigrade completes work more efficiently. `get_sources_cloudigrade_application_type_id` currently runs every time `create_from_sources_kafka_message` runs and results in another call to sources-api. We do not expect the returned value to change frequently, though it is _possible_ that it could change occasionally. The low rate of changes makes this likely a good candidate for caching or memoizing.

## Acceptance Criteria

- [ ] Verify `get_sources_cloudigrade_application_type_id` only calls sources-api once every X seconds as needed.
  - one hour TTL.
- [ ] Verify the value is not cached if sources-api doesn't return a reasonable value.
  - This may require us to cleverly nest a couple of function calls to raise an exception to prevent caching.
- [ ] Verify we have a way to reset/poison this cached value.
- [ ] Discover and share with the team knowledge about possible platform cache.
  - [ ] If this exists today, have our app use it for this value.
- [ ] All new environment variables are appropriately configurable through e2e-deploy/saas-templates/app-interface yamls

## Assumptions and Questions

- How long should we keep the cached value?
  - one hour
- We could naively memoize in-memory, and that reduce the API calls to one per worker per TTL. Or should we add a caching server (Redis?) to our OpenShift deployments that would be shared among all of our instances?
  - Have a discussion with someone on the platform team about caching servers. Like they provide databases and message queues, they should also provide standard shared caching tools. If this exists, share knowledge with the team and consider using it.
  - When we recently discussed this issue, Chris D suggested that we run our own Redis instance. That's what other teams are doing because there does not appear to be a platform standard caching server.
- Consider [django-cache-memoize](https://pypi.org/project/django-cache-memoize/) or [django-memoize](https://pypi.org/project/django-memoize/).

gtanzillo:
Leveraging `django-cache-memoize` to implement caching the value of `get_cloudigrade_application_type_id`, since application types

rarely changes on sources, and avoids making unnecessary calls to the Sources API. The cache ttl is 60 seconds here but it could easily be much higher

for #663

### DIFF
--- a/cloudigrade/config/settings/base.py
+++ b/cloudigrade/config/settings/base.py
@@ -496,3 +496,12 @@ SOURCES_AVAILABILITY_EVENT_TYPE = env(
     "SOURCES_AVAILABILITY_EVENT_TYPE",
     default="availability_status",
 )
+
+# Cache ttl settings
+CACHE_TTL_DEFAULT = env.int(
+    "CACHE_TTL_DEFAULT", default=60
+)
+
+CACHE_TTL_SOURCES_APPLICATION_TYPE_ID = env.int(
+    "CACHE_TTL_SOURCES_APPLICATION_TYPE_ID", default=CACHE_TTL_DEFAULT
+)
--- a/cloudigrade/util/redhatcloud/sources.py
+++ b/cloudigrade/util/redhatcloud/sources.py
@@ -4,6 +4,7 @@ import json
 import logging
 
 import requests
+from cache_memoize import cache_memoize
 from confluent_kafka import KafkaException, Producer as KafkaProducer
 from django.conf import settings
 from django.utils.translation import gettext as _
@@ -187,6 +188,7 @@ def extract_ids_from_kafka_message(message, headers):
     return account_number, platform_id
 
 
+@cache_memoize(60)
 def get_cloudigrade_application_type_id(account_number):
     """Get the cloudigrade application type id from sources."""
     sources_api_base_url = settings.SOURCES_API_BASE_URL
--- a/cloudigrade/util/redhatcloud/sources.py
+++ b/cloudigrade/util/redhatcloud/sources.py
@@ -188,7 +188,7 @@ def extract_ids_from_kafka_message(message, headers):
     return account_number, platform_id
 
 
-@cache_memoize(60)
+@cache_memoize(settings.CACHE_TTL_SOURCES_APPLICATION_TYPE_ID)
 def get_cloudigrade_application_type_id(account_number):
     """Get the cloudigrade application type id from sources."""
     sources_api_base_url = settings.SOURCES_API_BASE_URL
--- a/cloudigrade/util/tests/redhatcloud/test_sources.py
+++ b/cloudigrade/util/tests/redhatcloud/test_sources.py
@@ -133,6 +133,30 @@ class SourcesTest(TestCase):
         mock_get.assert_called()
 
     @patch("requests.get")
+    def test_get_sources_cloudigrade_application_type_is_cached(self, mock_get):
+        """Assert get_cloudigrade_application_type_id returns cached id."""
+        cloudigrade_app_type_id = _faker.pyint()
+        expected = {"data": [{"id": cloudigrade_app_type_id}]}
+        mock_get.return_value.status_code = http.HTTPStatus.OK
+        mock_get.return_value.json.return_value = expected
+
+        response_app_type_id = sources.get_cloudigrade_application_type_id(
+            self.account_number
+        )
+        self.assertEqual(response_app_type_id, cloudigrade_app_type_id)
+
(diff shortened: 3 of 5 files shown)


cloudigrade/util/tests/redhatcloud/test_sources.py
```diff
@@ -112,6 +112,30 @@ def test_get_sources_cloudigrade_application_type_success(self, mock_get):
...
+        """Call it again and make sure the original, cached value is returened"""
+        not_cloudigrade_app_type_id = _faker.pyint()
+        not_expected = {"data": [{"id": not_cloudigrade_app_type_id}]}
+        mock_get.return_value.json.return_value = not_expected
+
+        response_app_type_id = sources.get_cloudigrade_application_type_id(
+            self.account_number
+        )
+        self.assertEqual(response_app_type_id, cloudigrade_app_type_id)
+        mock_get.assert_called_once()
```
gtanzillo, line 137: @abellotti I don't think there's a convenient way to use `assert_called_once_with` because the argument is the request being made to Sources and is built in the `get_cloudigrade_application_type_id` function

infinitewarp:
FYI @gtanzillo I edited the description up top to say "for #663" instead of "Fixes #663" so that GitHub doesn't automatically close the issue, but then I noticed that it's actually in the commit message itself. Would you amend that commit to remove the "fixes" keyword and force push it here?

We generally prefer to leave the original issue open so someone else can verify all relevant acceptance criteria before closing.

infinitewarp (changes requested):
FYI @gtanzillo I edited the description up top to say "for #663" instead of "Fixes #663" so that GitHub doesn't automatically close the issue, but then I noticed that it's actually in the commit message itself. Would you amend that commit to remove the "fixes" keyword and force push it here?

We generally prefer to leave the original issue open so someone else can verify all relevant acceptance criteria before closing.

infinitewarp (dismissed):
FWIW, I started up a local docker container running `memcached` exposed on `11211`, temporarily `poetry add python-memcached` into my virtualenv, and added to my cloudigrade config:

```
CACHES = {
    'default': {
        'BACKEND': 'django.core.cache.backends.memcached.MemcachedCache',
        'LOCATION': '127.0.0.1:11211',
    }
}
```

…because I wanted to see if this `cache_memoize` decorator works as I hoped. It looks good to me! 🙂 When the `memcached` instance is running and responding on `11211`, each time I called a new decorated function, I got the same cached result after the first. As soon as I killed the server, my decorated function seamlessly started running "normally" with no cache step.

~~This also means that until we implement a real caching server, this change gives us no real improvement. So, while I'm okay approving this PR, that means we still have work ahead of us before we can call https://github.com/cloudigrade/cloudigrade/issues/663 complete.~~

Correction: once I remove the aforementioned `CACHES` config, it looks like Django or this library falls back to some kind of in-memory cache. So, I redact my earlier statement! This means the decorator _should_ result in the value being cached, even without the server, but it would still do us well to implement a real caching server at some point.

cloudigrade/util/redhatcloud/sources.py
```diff
@@ -182,6 +183,7 @@ def extract_ids_from_kafka_message(message, headers):
     return account_number, platform_id
 
 
+@cache_memoize(60)
```
katherine-black, line 186: Should we consider pulling the 60 here out into the configuration file? Would make it possible to adjust via an env var, so if we ever have to change it we can just update the deployment, not build a new image.

gtanzillo: Yes, that makes sense, I'll take care of that. Will WIP this PR in the meantime. Thanks!

katherine-black (dismissed):

cloudigrade/config/settings/base.py
```diff
@@ -496,3 +496,12 @@
...
 )
+
+# Cache ttl settings
+CACHE_TTL_DEFAULT = env.int(
+    "CACHE_TTL_DEFAULT", default=60
+)
+
+CACHE_TTL_SOURCES_APPLICATION_TYPE_ID = env.int(
+    "CACHE_TTL_SOURCES_APPLICATION_TYPE_ID", default=CACHE_TTL_DEFAULT
+)
```
katherine-black, line 507: Nitpick, but could we get an empty line at the end of this file? It helps with avoiding merge conflicts.

katherine-black (approved):

cloudigrade/config/settings/base.py
katherine-black: Thanks!

infinitewarp (approved):

Issue #907, opened 2021-05-17
more specific customer-facing error codes


infinitewarp:
## Summary

As a subscription watch customer, I want more useful error data from cloudigrade when my cloud meter source is "unavailable" so that I can correct the issue without contacting support.

We have several conditions in `verify_permissions` that all result in a generic `CG3000` error code going to sources-api. I think we could make some more specific error codes without risking leaked information about other customers to a malicious user. I propose we:

- make if clear that `CG3000` is a generic failure, and the underlying cause my not be known
- continue using `CG3001` to indicate that the AWS says the customer has too many trails
- add `CG3002` to indicate that we got AccessDenied when using the ARN, suggesting that the role was misconfigured
- add `CG3003` to indicate that we were able to use the ARN but failed to verify at least one policy action, suggesting that the policy is misconfigured
- add whatever others seem appropriate as we're refactoring this behavior

To be clear, this means that where we previously would send a generic `CG3000`, we might instead sometimes send more specific `CG3002` or `CG3003` (or `CG300x`?) codes if we have enough information to make that differentiation.

## Acceptance Criteria

- [x] Verify each of these new error codes appears in sources-api's `application.availability_status_error` from each use case

## Assumptions and Questions

- This is a followup to some `TODO` comments that @infinitewarp added with https://github.com/cloudigrade/cloudigrade/pull/908

infinitewarp:
@paraggit confirmed at today's standup that his testing shows the new codes and messages are being used.

Issue #892, opened 2021-05-06
sources is not updated when AWS role has insufficient permissions.


paraggit:
## Summary

This issue was originally reported by a customer. In this case the customer has created sources and its remains in the 'Unknown' state. To create a sources they followed a manual method in which policy and role is created by the customer as per the instructions given in the sources screen. From the investigation it got to know that an issue is caused due to AWS role has insufficient permission to operate cloudmeter application and hence cloudmeter is not proceed further to add account but this action is not intimated back to the sources application.

In some cases if there is an issue to add sources, cloudmeter is reporting back to the application object with the proper error code and message and these are defined [ here.](https://github.com/cloudigrade/cloudigrade/wiki/Error-Codes) . 

## Steps to Reproduce

1. Login to C.R.C and click on settings menu.
2. Click on sources menu.
3. Click on 'Add Sources' button.
4. Set sources name.
5. click on  'manual method' radio button.
6. on next screen copy the policy json and create policy on AWS but before that remove some permissions mentions in that policy.
7. add role in AWS with an account number mention in sources screen.
8. copy AWS ARN from AWS portal and past it to given text box.
9. Click on create sources button. 

## Expected Result

- Sources should created successfully and it has status as 'Active'

## Actual Result

- Sources is created but their status remains as 'Unknown' 

## Additional context

- @infinitewarp believes that this issue was the root cause for the recent (2021-05-05) customer-reported issue. Ask @infinitewarp if you want details.
- Here is the screen snippet of sources which stays as 'Unknown'
<img width="1459" alt="sources" src="https://user-images.githubusercontent.com/14196504/117311292-b3890080-aea1-11eb-818a-6fad899a10e4.png">

- Application object is not has the error message.
```shell

%  http  --proxy=$PROXY  --verify=no --auth $AUTH "https://$URL/applications/20312/"
HTTP/1.1 200 OK
Cache-Control: max-age=0, private, must-revalidate
Connection: keep-alive
Content-Encoding: gzip
Content-Length: 151
Content-Type: application/json; charset=utf-8
Date: Wed, 05 May 2021 09:51:57 GMT
ETag: W/"5bccdd4f82f059c5a191dd110202df84"
Server: openresty/1.13.6.1
Set-Cookie: 8f585a7d113e400d14faa100d81cbdaf=1be5c74193c334cb250bdbe9f5959d98; path=/; HttpOnly; Secure; SameSite=None
Vary: Origin
Vary: Accept-Encoding
[...]

PR #908, opened 2021-05-17
reorganize clouds.aws.util.verify_permissions

infinitewarp:
reorganize clouds.aws.util.verify_permissions

- always notify sources about errors encountered here
- be more explicit about handling different exceptions
- make unit tests cover happy and all major error paths
- stop raising ValidationError since only the seldom-used internal HTTP API cares about that
- stop notifying sources in verify_account_permissions since verify_permissions would have already done that

This also includes a minor fix to an old-style deprecated `task` import that should be `shared_task`.

for https://github.com/cloudigrade/cloudigrade/issues/892

Note that I added `TODO`s in some comments about adding more error codes. I opened the following issue for that: https://github.com/cloudigrade/cloudigrade/issues/907

### DIFF
--- a/cloudigrade/api/clouds/aws/models.py
+++ b/cloudigrade/api/clouds/aws/models.py
@@ -86,17 +86,11 @@ class AwsCloudAccount(BaseModel):
         """
         from api.clouds.aws import tasks, util  # Avoid circular import.
 
-        try:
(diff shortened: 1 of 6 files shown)


katherine-black (approved):
Good change 👍

Issue #723, opened 2021-01-29
Define recurring tasks in migrations, not configs


katherine-black:
In GitLab by @infinitewarp on Aug 19, 2020, 14:44

## Summary

As a cloudigrade sysop, I want the definition of periodic Celery tasks that live in the database to be more clearly defined and predictably created and updated. Currently, periodic Celery tasks are created and updated automatically by being defined in the `CELERY_BEAT_SCHEDULE` dict in our Django settings. This story intends to move that dict's definitions into Django DB migrations.

## Acceptance Criteria

- [ ] Verify old scheduled tasks still exist with the same configuration
- [ ] Verify the new process supports some method of easily overriding the schedule timing

## Assumptions and Questions

- One of the benefits of keeping periodic Celery task definitions in the `CELERY_BEAT_SCHEDULE` settings is that this provides a very convenient mechanism to change the schedule in different environments. We often need to override the defaults especially in the CI and QA environments to facilitate testing and verification. Whatever new solution we implement also needs to easily accept something like environment variables to update the schedule.
  - How do we do this with migrations?
    - Brad's Bad Idea #1: Write some code. Things in base.py that read vars and recreate celery functionality for dict handling.
    - Brad's Bad Idea #2, Electric Boogaloo: Manage command to do this.
      - Kat's Compilation Idea: Management command defines set of tasks with set IDs. On deployment, same time we run our migrations, we torch ALL tasks with those specific IDs. We recreate the tasks from the management command. 
        - Brad's: Upsert not delete.
    - Kat's Bad Idea #1: Write some _conditional_ migrations. What does this mean...? If CI RUN extra migrations..?
      - Specific CI/QA setup. Test Admin Accounts + other garbage???

katherine-black:
Logs from a beat bot in QA

```
2021-05-19 17:23:40,483 \| INFO \| 1 \| schedulers.py:schedule:356 \| requestid:None \| DatabaseScheduler: Schedule changed.
--
2021-05-19 17:23:41,034 \| INFO \| 1 \| beat.py:apply_entry:267 \| requestid:None \| Scheduler: Sending due task analyze_log_every_2_mins (api.clouds.aws.tasks.analyze_log)
--
2021-05-19 17:24:10,596 \| INFO \| 1 \| beat.py:apply_entry:267 \| requestid:None \| Scheduler: Sending due task analyze_log_every_2_mins (api.clouds.aws.tasks.analyze_log)
--
[...]

PR #909, opened 2021-05-18
Turn env var periodic task definition into programmatic definition as…

katherine-black:
… part of celery startup.

#723

### DIFF
--- a/cloudigrade/config/celery.py
+++ b/cloudigrade/config/celery.py
@@ -22,6 +22,40 @@ logger.info("Django setup.")
 
 app = Celery("config")
 app.config_from_object("django.conf:settings", namespace="CELERY")
+# Remember: the "schedule" values are integer numbers of seconds.
+app.conf.beat_schedule = {
+    "delete_inactive_users": {
+        "task": "api.tasks.delete_inactive_users",
+        "schedule": env.int("DELETE_INACTIVE_USERS_SCHEDULE", default=24 * 60 * 60),
+    },
+    "persist_inspection_cluster_results": {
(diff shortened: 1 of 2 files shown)


abellotti (approved):

infinitewarp (changes requested):
This seems reasonable at a glance, but I'm not closely familiar with the underlying mechanism that loads this. Does this mean that the beat schedule is now exclusively defined by config, foregoing the database model entirely? If so, should we delete the old scheduled tasks to make extra sure that they don't get picked up or conflict?

Also, have you tried running this locally, stopping to change some schedule values, and restart to see if they update as expected?

katherine-black:
> This seems reasonable at a glance, but I'm not closely familiar with the underlying mechanism that loads this. Does this mean that the beat schedule is now exclusively defined by config, foregoing the database model entirely? If so, should we delete the old scheduled tasks to make extra sure that they don't get picked up or conflict?

No, nothing here changes at all except how we define these tasks. I can not explain why one method works and the other does not, they both technically did the same thing. My guess is there is a bug with how Celery translates certain env var config values into actual configuration parameters, especially for data that was already defined. But yes...

> Also, have you tried running this locally, stopping to change some schedule values, and restart to see if they update as expected?

I did in fact run this locally. Starting beat up, letting it set periodic tasks in the db. Shut it down, updated an env var with a specific time frame for an already set task, upon beat start up, a new interval entry was added to that table and the existing task updated to reference the new interval.

gtanzillo (approved):

infinitewarp (approved):

PR #924, opened 2021-06-01
Adding support for the x-rh-internal-fake-identity header for the internal concurrent API

abellotti:
Adding support for the account_number and org_admin headers for the internal/concurrent API.
    
when hitting the internal concurrent API, the x-rh-identity header for the associate does not include the account_number or the is_org_admin flag, this enhancement allows us to optionally specify those in the headers:
- X_RH_IDENTITY_ACCOUNT
- X_RH_IDENTITY_ORG_ADMIN

Marking as WIP:
- [x] Needs additional internal concurrent tests
- [x] Shift to headers instead of parameters, current parameter account number and is_org_admin persist in response links and such, all auth/identity should stay inbound in headers or cookies. Also, this way we're consistent with x-rh-identity, so add support for optional headers instead:
    - x-rh-identity-account
    - x-rh-identity-org-admin
- [x] Shift to using a single internal fake insights identity header X-RH-INTERNAL-FAKE-IDENTITY
 
Additional enhancement needed for https://github.com/cloudigrade/cloudigrade/issues/894


cloudigrade/api/authentication.py
```diff
@@ -95,6 +96,8 @@ class IdentityHeaderAuthentication(BaseAuthentication):
     require_account_number = True
     require_user = True
     create_user = False
+    support_account_number_header = False
+    support_org_admin_header = False
```
infinitewarp, line 100: Rather than take in these two specific values, I think we could better future-proof this by having one new header that we expect to have the same structure/content as the normal identity header. Then we can simplify the change in `parse_requests_header` below simply to load one header value over the other.

```suggestion
    allow_internal_fake_identity_header = False
```

cloudigrade/api/authentication.py
```diff
@@ -175,6 +178,22 @@ def authenticate(self, request):
...
+                logger.info(_("Using account_number '%s' from header"), account_number)
+
+        if self.support_org_admin_header:
+            org_admin_header = request.META.get(
+                settings.INSIGHTS_IDENTITY_ORG_ADMIN_HEADER, None
+            )
+            if org_admin_header:
+                is_org_admin = BooleanField().to_python(org_admin_header)
+                logger.info(_("Using is_org_admin '%s' from header"), is_org_admin)
+
```
infinitewarp, line 196: Per the comment above, I'd suggest removing all of this change. 

Instead, in `parse_requests_header`, where we currently have this:

```python
auth_header = request.META.get(settings.INSIGHTS_IDENTITY_HEADER, None)
```

…I think that could become this:
```python
auth_header = request.META.get(settings.INSIGHTS_IDENTITY_HEADER, None)
if not auth_header and self.allow_internal_fake_identity_header:
    auth_header = request.META.get(settings.INTERNAL_FAKE_INSIGHTS_IDENTITY_HEADER, None)
```

This means we would always try to load the standard header first. If we don't get one _and_ we allow overriding it, then we load the new internal header.

This assumes that we normally get nothing in the standard header for internal requests. I don't know if that's actually the case or if we get something different or partial. My suggestion probably is too simple and needs a little more conditional help.

Maybe we flip the order? If the "allow" flag is true, we could load the new header first, and then if it's still empty, we load the standard header.

```suggestion
```

cloudigrade/config/settings/base.py
```diff
@@ -488,6 +488,8 @@ def __print(*args, **kwargs):
 CLOUDIGRADE_REQUEST_HEADER = "X-CLOUDIGRADE-REQUEST-ID"
 INSIGHTS_REQUEST_ID_HEADER = "HTTP_X_RH_INSIGHTS_REQUEST_ID"
 INSIGHTS_IDENTITY_HEADER = "HTTP_X_RH_IDENTITY"
+INSIGHTS_IDENTITY_ACCOUNT_HEADER = "HTTP_X_RH_IDENTITY_ACCOUNT"
+INSIGHTS_IDENTITY_ORG_ADMIN_HEADER = "HTTP_X_RH_IDENTITY_ORG_ADMIN"
```
infinitewarp, line 492: I think these new names looks too close to the real identity header name. It should be abundantly clear that we're (sometimes) paying attention to extra headers that fake or override the primary identity header.

Here's one possible suggestion that collapses the two into one (as mentioned elsewhere in my comments):

```suggestion
INTERNAL_FAKE_INSIGHTS_IDENTITY_HEADER = "HTTP_X_RH_INTERNAL_FAKE_IDENTITY"
```

cloudigrade/internal/authentication.py
```diff
@@ -35,3 +35,19 @@ class IdentityHeaderAuthenticationInternalCreateUser(IdentityHeaderAuthenticatio
     require_account_number = True
     require_user = False
     create_user = True
+
+
+class IdentityHeaderAuthenticationInternalConcurrent(IdentityHeaderAuthentication):
```
infinitewarp, line 40: I like this idea so much that I think we could use it in more internal APIs! Maybe not right now but in the near future. So, rather than tie this name specifically to the concurrent API, how about something like this?

```suggestion
class IdentityHeaderAuthenticationInternalFakeIdentity(IdentityHeaderAuthentication):
```

Maybe that's not the best name either… but something along those lines?

cloudigrade/internal/authentication.py
```diff
@@ -35,3 +35,19 @@ class IdentityHeaderAuthenticationInternalCreateUser(IdentityHeaderAuthenticatio
...
+    This authentication checks for the identity header and requires the identity
+    to have org_admin enabled.
+    """
+
+    require_org_admin = True
+    require_account_number = True
+    require_user = True
+    create_user = False
+    support_account_number_header = True
+    support_org_admin_header = True
```
infinitewarp, line 53: …and if we collapse the two as I mentioned above, this would be:
```suggestion
    allow_internal_fake_identity_header = True
```

infinitewarp (changes requested):
Overall I like this idea of conditionally looking for a new header! I have one general request, though, that is probably repeated and scattered through the inline comments: let's make this more general-purpose and future-proof with something like the following:

- Use a request header name that is _very obviously_ not the main header so nobody would confuse them at a glance.
- Instead of two new headers specifically for account number and org admin, use one new header that expects the same structure/content as the existing identity header. This should simplify the current diff, and if we start using more pieces of the main identity header, we won't have to keep adding new attribute-specific headers and conditions.
- Use a class name that we can more easily use in other internal APIs without needing to rename it or make more versions of the same.

cloudigrade/api/authentication.py
```diff
@@ -13,7 +13,7 @@
 logger = logging.getLogger(__name__)
 
 
-def parse_requests_header(request):
+def parse_requests_header(self, request):
```
infinitewarp, line 16: Although it isn't a reserved word, strongly by convention `self` is only used class instance methods to refer to the instance. Since this is a class-less function, this should be named something else more descriptive of the incoming object type.

infinitewarp, line 16: Actually, on second look, since we really only use the `.allow_internal_fake_identity_header` attribute of the new object being passed in, maybe we should just be more precise like:

```suggestion
def parse_requests_header(request, allow_internal_fake_identity_header=False):
```

…and change the caller to send only that attribute, not its whole self.

cloudigrade/api/authentication.py
```diff
@@ -32,6 +32,10 @@ def parse_requests_header(request):
     )
 
     auth_header = request.META.get(settings.INSIGHTS_IDENTITY_HEADER, None)
+    if self.allow_internal_fake_identity_header:
```
infinitewarp, line 35: Do we _always_ want to override the standard header if we're allowing the internal fake identity header? This is where my knowledge of the internal Turnike use case is a bit thin. Does Turnpike send anything at all in the position of the regular identity header? Or is it safe to just discard it entirely in favor of our fake one?

abellotti: Yes, this is fine. the only 2 things we care about are the account_number and is_org_admin which do not exist in the Turnpike generated header anyways.  When we communicate internally to sources we generate the identity header ourselves using the account number we obtained, in other words we never forward the whole header as is downstream so we really don't use any other info from the header.

infinitewarp (approved):

Issue #918, opened 2021-05-26
Improve alerting around KafkaProducer failures


infinitewarp:
## Summary

As a cloudigrade sysop, I want to be notified promptly of when KafkaProducer fails unexpectedly so I can react to fix things. We recently had incorrect configs in our cloudigrade-worker deployments that resulted in an invalid Kafka hostname. However, when we used KafkaProducer with that invalid hostname, we were not alerted to any errors.

@infinitewarp checked the cloudigrade-worker pods logs directly during this incident and found messages like this in stage:

```
2021-05-25 18:46:16,601 | INFO | 9 | sources.py:notify_application_availability:238 | requestid:ecceca7b-37b0-47b0-a596-eaf74a7cddda | Requesting the update of the availability status for application 25485 as unavailable
2021-05-25 18:46:16,601 | INFO | 9 | sources.py:notify_application_availability:248 | requestid:ecceca7b-37b0-47b0-a596-eaf74a7cddda | Instantiating KafkaProducer with {'bootstrap.servers': 'platform-mq-ci-kafka-bootstrap.platform-mq-ci.svc:9092'}
2021-05-25 18:46:16,602 | INFO | 9 | sources.py:notify_application_availability:259 | requestid:ecceca7b-37b0-47b0-a596-eaf74a7cddda | KafkaProducer will produce with topic='platform.sources.status' value='{"resource_type": "Application", "resource_id": 25485, "status": "unavailable", "error": "Could not enable cloud metering. Please contact support. Error code CG3000."}' headers='{'event_type': 'availability_status'}'
2021-05-25 18:46:16,602 | INFO | 9 | sources.py:notify_application_availability:276 | requestid:ecceca7b-37b0-47b0-a596-eaf74a7cddda | KafkaProducer produced!
%3|1621968376.607|FAIL|rdkafka#producer-10| [thrd:platform-mq-ci-kafka-bootstrap.platform-mq-ci.svc:9092/bootstra]: platform-mq-ci-kafka-bootstrap.platform-mq-ci.svc:9092/bootstrap: Failed to resolve 'platform-mq-ci-kafka-bootstrap.platform-mq-ci.svc:9092': Name or service not known (after 5ms in state CONNECT)
2021-05-25 18:46:17,095 | INFO | 1 | strategy.py:task_message_handler:151 | requestid:None | Received task: api.tasks.notify_application_availability_task[cafe1cd6-9786-4a5e-ab5b-2afe7d778740]  
[...]

PR #926, opened 2021-06-08
Add callback function to verify delivery of messages to kafka.

katherine-black:
Related to #918


abellotti:
with the addition of a callback, do we need to add back the poll(0) to make sure we get them all as was done here https://github.com/cloudigrade/cloudigrade/pull/890/commits/608532678a699376734210ca9d6a3abeb110e744#diff-623f27daa14bc29755606a95e78f44026a95c0404a8d99b8d8573753646e2d7bL249  ?  I had removed it because it had no effect without a callback.

katherine-black:
> with the addition of a callback, do we need to add back the poll(0) to make sure we get them all as was done here [6085326#diff-623f27daa14bc29755606a95e78f44026a95c0404a8d99b8d8573753646e2d7bL249](https://github.com/cloudigrade/cloudigrade/commit/608532678a699376734210ca9d6a3abeb110e744#diff-623f27daa14bc29755606a95e78f44026a95c0404a8d99b8d8573753646e2d7bL249) ? I had removed it because it had no effect without a callback.

No, flush will handle calling it. There is no explicit mention of it in their docs, but its [here](https://github.com/confluentinc/examples/blob/6.1.1-post/clients/cloud/python/producer.py#L47) in their examples.

It also looks like it didn't have any effect previously in your diff because it needs to be after the produce call.

cloudigrade/util/redhatcloud/sources.py
```diff
@@ -296,3 +297,19 @@ def notify_application_availability(
...
+    if error is not None:
+        logger.error(
+            _("Failed to deliver message: %(message)s due to: %(error)s"),
+            {"message": message, "error": error},
+        )
+    else:
+        log_msg = _("Successfully delivered message: %(message)s"), {"message": message}
+
+        if settings.VERBOSE_SOURCES_NOTIFICATION_LOGGING:
+            logger.info(log_msg)
```
infinitewarp, line 313: Does this actually work as expected? I would naively expect we'd need to explode the `log_msg` tuple like 

```suggestion
            logger.info(*log_msg)
```

I tried something similar in a local shell:

```
>>> log_msg = "hello %(foo)s", {"foo": "world"}
>>> logger.info(log_msg)
2021-06-09 18:35:37,496 | INFO | 38488 | <console>:<module>:1 | requestid:None | ('hello %(foo)s', {'foo': 'world'})
>>> logger.info(*log_msg)
2021-06-09 18:35:41,029 | INFO | 38488 | <console>:<module>:1 | requestid:None | hello world

infinitewarp (approved):

PR #933, opened 2021-06-22
improve local dev experience

infinitewarp:
This PR adds django-extensions to improve local experimentation, updates our non-prod settings modules to behave more like prod, and updates some woefully out-of-date developer setup instructions in the README.

In more detail:

- add django-extensions, jupyter, ipython as dev dependencies to improve local code experimentation
- update readme to remove old irrelevant notes about shiftigrade and local openshift commands
- update readme to correct commands for updating rest api examples doc and openapi.json files
- update readme to explain how to use `ansible-notebook` for AWS setup
- update `base.py` to use `CLOUDIGRADE_ENVIRONMENT` like `prod.py`
- update readme to explain the current set of preferred environment variables
- remove redundant settings in `prod.py` now that `base.py` uses `CLOUDIGRADE_ENVIRONMENT`
- replace irrelevant travis-ci badge with current github actions badge

Since the `README` diff is kind of huge, it might be easier to just look at this branch's version directly here: https://github.com/cloudigrade/cloudigrade/blob/better-dev-experience/README.rst In particular, the "AWS account setup" and "Environment variables" sections have been heavily rewritten.

The `base.py` diff is also kind of huge, and it might also be easier to look at this branch's version directly here: https://github.com/cloudigrade/cloudigrade/blob/better-dev-experience/cloudigrade/config/settings/base.py When I was updating variables to read from env vars more like prod, I also *moved around* large sections of the file and added inline comment `####` dividers to try to give it (hopefully) a more readable, organized structure. I was tempted to actually split it into multiple separate modules, with different modules focusing on a specific area of interest, but I decided just to try grouping with comments for now. I'm curious if anyone has further thoughts on that kind of approach since we have lots of configs.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -277,20 +277,20 @@ You may run the following Make target to generate the API examples documentation
 
 This will create many use-case-specific records in the database, simulate API calls through cloudigrade, and generate an updated document with the API calls. You should review any changes made by this command before adding and committing them to source control.
 
-Generate a Spec File
---------------------
+Generate openapi.json Files
+---------------------------
 
-Generation of the spec file is handled by the same mechanism that serves our spec file via api, to ensure that they are the same. If you've recently made changes to the api and need to update the spec file, run the following command:
-
-.. code-block:: sh
-
-    make spec
-
-Otherwise, if you'd simply like to verify that the spec is current, you can run the following:
+Generation of the ``openapi.json`` and ``openapi-internal.json`` files uses the same mechanism that dynamically serves the specifications via the API, and the static files' contents should always match what the API serves dynamically. If you've recently made changes to the API and need to update the static files, run the following command:
 
 .. code-block:: sh
 
-    make spec-test
+    make openapi
+
+Otherwise, if you'd simply like to verify that the current static files match the API, you can run the following command:
+
+.. code-block:: sh
+
+    make openapi-test
 
 
 Authentication
--- a/README.rst
+++ b/README.rst
@@ -269,13 +269,7 @@ If you wish to run a higher-level suite of integration tests, see `integrade <ht
 Updating API Example Docs
 -------------------------
 
-To automatically update the API examples documentation, you need a database with current migrations applied but with no customer data in it.
-
-.. code-block:: sh
-
-    make oc-forward-ports
-
-Once the database is available, you may run the following Make target to generate the API examples documentation:
+You may run the following Make target to generate the API examples documentation:
 
 .. code-block:: sh
 
--- a/README.rst
+++ b/README.rst
@@ -2,7 +2,7 @@
 cloudigrade
 ***********
 
-|license| |Build Status| |codecov| |Updates| |Python 3|
+|license| |Build Status| |codecov|
 
 
 What is cloudigrade?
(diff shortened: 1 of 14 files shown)


README.rst
```diff
@@ -354,53 +263,31 @@ If you wish to run *only* the tests:
...
 -------------------------
 
-To automatically update the API examples documentation, you need a database with current migrations applied but with no customer data in it. If you have deployed to a local OpenShift cluster, you should forward the database port so it can be accessed locally.
-
-.. code-block:: sh
-
-    make oc-forward-ports
-
-Once the database is available, you may run the following Make target to generate the API examples documentation:
+Tou may run the following Make target to generate the API examples documentation:
```
katherine-black, line 274:
```suggestion
You may run the following Make target to generate the API examples documentation:
```

Typo.

cloudigrade/config/settings/base.py
```diff
@@ -373,44 +396,25 @@ def __print(*args, **kwargs):
...
+#####################################################################
+# Celery broker
 
 CELERY_BROKER_TRANSPORT_OPTIONS = {
     "queue_name_prefix": AWS_NAME_PREFIX,
     "region": AWS_SQS_REGION,
 }
 CELERY_BROKER_URL = AWS_SQS_URL
-QUEUE_EXCHANGE_NAME = None
+QUEUE_EXCHANGE_NAME = None  # TODO Does anything still use this? Please comment.
```
katherine-black, line 408:
```suggestion
```
Should be safe to delete, this setting is not referenced anywhere.

cloudigrade/config/settings/prod.py
```diff
@@ -9,23 +9,16 @@
 
 DJANGO_DEBUG = "False"
 
+# CLOUDIGRADE_ENVIRONMENT may have been loaded in base.py, but if not present, it would
+# have fallen back to generic defaults. However, in prod, we *require* that it be set
+# explicitly. By reloading here *without* a default, we halt app startup if not present.
 CLOUDIGRADE_ENVIRONMENT = env("CLOUDIGRADE_ENVIRONMENT")
 
+# TODO Why isn't HOUNDIGRADE_AWS_AVAILABILITY_ZONE configurable? Why is it special here?
```
katherine-black, line 17: Because of this: https://github.com/cloudigrade/cloudigrade/blob/master/deployment/playbooks/roles/aws/defaults/main.yml#L40

But more realistically because long ago we agreed to setup and run everything from us-east-1a (since our 1a is not the same as everyone elses 1a), and as infra was manual setup, configs were not made to be easily changeable as we didn't expect a lot of change there.

We can go ahead and change this to pull from an env var, but then we should also update the playbook to source it from the same env var, and now the number of playbook inputs grows from 1 to 2 🙂 

Let me know what you think, I'm definitely open to making this better.

katherine-black: Cloud-init is another spot where this will likely change.

infinitewarp: Good points! I'll add an inline comment summarizing your comment here, and I'll leave it to "future us" to refactor or remove later.

katherine-black (approved):

Issue #729, opened 2021-01-29
Populate Instance Definitions for Azure


katherine-black:
In GitLab by @werwty on Sep 1, 2020, 14:03

## Summary

For AWS, we query their API priodically to map all available AWS EC2 Instance Types to their memory and vcpu counts. We need to do the same for Azure.

## Acceptance Criteria

- [x] Verify a periodic task exists to get azure instance definitions and populate the InstanceDefinition table with this information
- [x] Verify that `manage.py loadec2defintions` load both AWS and Azure instance definitions
  - Rename that command since it's no longer only "EC2" definitions

## Assumptions and Questions

- Is there even an API for this? Something that can get a global list of options, not just "what I have available to me"?
- We might need to reach out to our MS/Azure contacts for help here.
- If we can't find a global list via API, have a conversation with the team to figure out next alternatives. :slight_smile: 
- Nothing for QE to do at this point. This work is still "in the black box", and its effects will be verified later when we look at the calculated output of the concurrent API.

katherine-black:
In GitLab by @infinitewarp on Sep 9, 2020, 16:21

changed title from **Populate Instace Definitions for Azure** to **Populate Insta{+n+}ce Definitions for Azure**

infinitewarp:
Confirmed we now have Azure definitions in prod. Thanks @katherine-black !

Example internal API call https://internal.cloud.redhat.com/api/cloudigrade/internal/api/cloudigrade/v1/instancedefinitions/397/

```
{
    "id": 397,
    "created_at": "2021-07-19T20:38:14.116444Z",
    "updated_at": "2021-07-19T20:38:14.116466Z",
    "instance_type": "Standard_B1ls",
    "memory_mib": 476,
    "vcpu": "1.00",
[...]

PR #937, opened 2021-06-28
729 azure definitions

katherine-black:
#729


cloudigrade/api/util.py
```diff
@@ -771,3 +773,52 @@ def process_instance_event(event):
             run.save()
             runs.append(run)
         calculate_max_concurrent_usage_from_runs(runs)
+
+
+def save_instance_type_definitions(definitions, cloud_type=AWS_PROVIDER_STRING):
```
infinitewarp, line 778: I like the decision to move this function.

I think `cloud_type` should become a required argument without a default, though, since AWS shouldn't be assumed as the primary caller, especially since we'll eventually add more clouds after Azure. I'm sure this function will evolve a lot more by the time we add GCP, but I think it sets a good precedent to *start* making the cloud types more explicit everywhere.

cloudigrade/api/clouds/azure/tasks/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+Important notes for developers:
+
+If you find yourself adding a new Celery task, please be aware of how Celery
+determines which queue to read and write to work on that task. By default,
+Celery tasks will go to a queue named "celery". If you wish to separate a task
+onto a different queue (which may make it easier to see the volume of specific
+waiting tasks), please be sure to update all the relevant configurations to
+use that custom queue. This includes CELERY_TASK_ROUTES in config and the
+Celery worker's --queues argument (see deployment-configs.yaml in shiftigrade
+and related configs in e2e-deploy and saas-templates).
```
infinitewarp, line 16: I figure this was copy-pasted from the other similar task modules. FWIW, the recently updated version of this text in those modules reads:
```suggestion
Celery worker's --queues argument (see related openshift deployment config files
elsewhere and in related repos like e2e-deploy and saas-templates).
```
Some day we should consolidate and clean up these notes, but I think duplicating them like this is fine for now.

cloudigrade/api/models.py
```diff
@@ -875,8 +875,8 @@ class InstanceDefinition(BaseModel):
     instance_type = models.CharField(
         max_length=256, null=False, blank=False, db_index=True
     )
-    memory = models.IntegerField(default=0)
-    vcpu = models.IntegerField(default=0)
+    memory = models.DecimalField(max_digits=32, decimal_places=2, default=0.00)
+    vcpu = models.DecimalField(max_digits=32, decimal_places=2, default=0.00)
```
infinitewarp, line 879: Out of curiosity, have you seen actual fractional CPU (and memory, I suppose) values come out of the Azure API? Or is this driven because their API docs specify these as float/decimal precision values?

cloudigrade/api/tests/clouds/azure/tasks/maintenance/test_repopulate_azure_instance_definitions.py
```diff
@@ -0,0 +1,114 @@
...
+        )
+
+        tasks.repopulate_azure_instance_mapping()
+
+        mock_compute.assert_called_once()
+        resource_skus.list.assert_called_once()
+        mock_db_create.assert_called_with(
+            instance_type="Standard_A5",
+            cloud_type="azure",
+            defaults={"memory": 14, "vcpu": 2, "json_definition": self.good_json},
```
infinitewarp, line 76: Since we now have higher decimal precision on how we store the memory and vcpu values, how would you feel about changing this or adding another test to exercise that precision? Might be unnecessary; just a passing thought.

cloudigrade/api/tests/clouds/azure/tasks/maintenance/test_repopulate_azure_instance_definitions.py
```diff
@@ -0,0 +1,114 @@
...
+            self.bad_standard_a5_definition, default=lambda x: x.__dict__
+        )
+
+        self.garbage_definition = Mock()
+        self.garbage_definition.name = "RandomAzureObject"
+        self.garbage_definition.resource_type = "notVirtualMachines"
+
+    @patch("api.util.InstanceDefinition.objects.get_or_create")
+    @patch(f"{_MAINTENANCE}.ComputeManagementClient")
+    def test_repopulate_azure_instance_mapping(self, mock_compute, mock_db_create):
```
infinitewarp, line 58: Why did you choose to mock out the interaction with the `InstanceDefinition` model? In most (all?) of our other tests, we let the DB write happen, and we query to see that it was written correctly. Is there something different about this particular test interaction that I might not be seeing?

cloudigrade/config/settings/base.py
```diff
@@ -55,6 +55,12 @@ def __print(*args, **kwargs):
 DEBUG = env.bool("DJANGO_DEBUG", default=False)
 ALLOWED_HOSTS = env("DJANGO_ALLOWED_HOSTS", default=["*"])
 
+# Azure Settings
+AZURE_CLIENT_ID = env("AZURE_CLIENT_ID", default="azure-client-id")
+AZURE_CLIENT_SECRET = env("AZURE_CLIENT_SECRET", default="very-secret-much-secure")
+AZURE_SUBSCRIPTION_ID = env("AZURE_SUBSCRIPTION_ID", default="azure-subscription-id")
+AZURE_TENANT_ID = env("AZURE_TENANT_ID", default="azure-tenant-id")
```
infinitewarp, line 62: Don't forget to add these to the README!

cloudigrade/api/util.py
katherine-black: I was on the fence with this, I appreciate the direction.

cloudigrade/api/models.py
katherine-black: Yea, Azure has 0.5 and such values. Worse yet, AWS reports memory in MB, while Azure in GB. We're storing them in the same field.

cloudigrade/api/tests/clouds/azure/tasks/maintenance/test_repopulate_azure_instance_definitions.py
katherine-black: No other reason than copying the pattern from `test_repopulate_ec2_instance_mapping.py`. I could rewrite both tests to remove the mock.

cloudigrade/api/models.py
katherine-black: We have an aws/azure flag so we can change how we interpret those values during calculation but that feels…. not great.

infinitewarp: Yikes! I thought I remembered AWS reporting in GB, but I just checked some real data and confirmed that, yes, the values we stored here are in MB.

This definitely sounds like an area we should refactor before too long. Maybe we could rename the vague `memory` to include units (`memory_mb`?) and multiply/divide as needed from the cloud-specific versions. Would you add a `# TODO` comment somewhere nearby to indicate that suggestion for future us? Or are you feeling ambitious enough to tackle that issue now?

katherine-black: Honestly, this will bite us if we let it slip for too long. I'm going to go ahead and make everything one unit so we don't end up ruining any future reports.

cloudigrade/api/tests/clouds/azure/tasks/maintenance/test_repopulate_azure_instance_definitions.py
infinitewarp: Huh. Not sure why we did that originally. 🤔 

Yeah, I think letting this go through the real model classes is the better path.

cloudigrade/api/models.py
katherine-black: fyi re: vcpu decimals
image

README.rst
```diff
@@ -189,6 +189,64 @@ If you want to undo that operation and effectively *remove* everything the playb
...
+                "actions": [
+                    "Microsoft.Compute/skus/read"
+                ],
+                "notActions": [],
+                "dataActions": [],
+                "notDataActions": []
+            }
+        ]
+
+#. Click `Review + create` -> `Create`.
```
infinitewarp, line 241: This is a pretty trivial point, but I figured I'd mention in case you didn't notice. If you intended for this number list to continue counting where it stopped before the code block (instead of resetting back to 1), *I think* the code block needs to be indented to make the parser think it's part of the previous `<li>`'s content.

I was looking at the formatted version here: https://github.com/cloudigrade/cloudigrade/blob/5b883be4cddc6f1a510bea8a1fe4738085b40a43/README.rst#azure-account-setup

It's really not a big deal, and I wasn't even sure about the intent. Feel free to disregard and resolve this comment if you like. 🙂

infinitewarp (approved):

infinitewarp (approved):

PR #950, opened 2021-07-16
Enabling Clowder for  Cloudigrade

abellotti:
Clowderizing Cloudigrade supporting deployment to minikube as well as the ephemeral cluster.

TODO:
- [x] Add the Azure config and parameters
- [x] clowderize and add support for postigrade in minikube/ephemeral
- [x] re-test on ephemeral once https://github.com/RedHatInsights/clowder/pull/389 is merged.


Dockerfile
```diff
@@ -28,12 +28,20 @@ RUN microdnf update \
     && microdnf remove libcurl-devel gcc python3-devel openssl-devel annobin -y \
     && microdnf clean all
 
+# Need jq for parsing the clowder configuration json in shell scripts
+RUN curl -L https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 -o jq \
```
katherine-black, line 32: Could we grab this from dnf/epel?

abellotti: yep, microdnf ftw!  👍

katherine-black: Thanks!

Dockerfile
```diff
@@ -28,12 +28,18 @@ RUN microdnf update \
     && microdnf remove libcurl-devel gcc python3-devel openssl-devel annobin -y \
     && microdnf clean all
 
+# Need jq for parsing the clowder configuration json in shell scripts
+RUN microdnf install jq -y \
```
infinitewarp, line 32: Any reason not to include this with the other packages above?

abellotti: The dnf installs seemed to be categorized, second one postgres, third one python.  I could add it to the first group though, seems general enough.

katherine-black (approved):

Issue #948, opened 2021-07-07
Simplify run and concurrent usage calculation


infinitewarp:
## Summary

As a cloudigrade developer and sysop, I want the calculations of runs and concurrent usage data to be only as complicated as necessary so that they produce correct data on a predictable schedule.

I am proposing we scrap the current event-driven mix of sync and async calculation functions and instead implement these calculations purely on a schedule. Consider the following possible workflow:

- cloudigrade records events during periodic cloud account describes
- cloudigrade records events whenever cloud trail provides them
- For each cloud account on a regular schedule (hourly?), for each instance in each account, if any events exist with a created at or occurred at date more recent than the last recorded run, recalculate runs since the oldest relevant event.
- For each cloud account on a regular schedule (hourly?), if any run exists with a created at date more recent than the last recorded concurrent usage, recalculate concurrent usage since the oldest relevant run.
- When a user requests concurrent usage data, simply return what is recorded.

There are a few possible benefits to an approach like this:

- Calculation of derivative denormalized data is decoupled from the actions that create or read that data. This means that a user cannot possibly spam up lots of tasks/calculations by hitting the api (or we don't need extra logic to protect or rate limit). It also means that very active accounts cannot possibly spam up or overlap tasks/calculations.
- If we simplify this as "for every cloud account" we can simplify and remove the current per-account scheduled tasks and the logic that ensures they exist. This may introduce a thundering herd problem, but we could have the main periodic scheduled task spawn another task for each account and/or give a slight jitter by using Celery's [the countdown argument](https://docs.celeryproject.org/en/stable/userguide/calling.html#eta-and-countdown).

## Acceptance Criteria

- [ ] Verify

## Assumptions and Questions

-

infinitewarp:
demo recording of the updated run and concurrent usage calculation process: https://asciinema.org/a/427887

infinitewarp:
@paraggit explained today at standup (2021-08-23) that he's verified concurrent usage calculations are now correct in stage. Thanks!

PR #951, opened 2021-07-19
simplify run and concurrent usage calculations

infinitewarp:
for https://github.com/cloudigrade/cloudigrade/issues/948

These commits should result in the following changes:

- run objects are no longer automatically rebuilt when we generate new events
- concurrentusage tasks are no longer automatically created when we generate new runs
- run calculation logic is simplified (likely fixing some preexisting bugs)
- run objects are created as needed on a periodic task for each user
- concurrentusage objects are created as needed on a periodic task for each user


cloudigrade/api/tests/util/test_get_max_concurrent_usage.py
```diff
@@ -69,10 +44,10 @@ def test_single_rhel_run_result(self, mock_schedule_concurrent_calc):
     @util_helper.clouditardis(util_helper.utc_dt(2019, 5, 3, 0, 0, 0))
     def test_single_rhel_run_no_result(self):
         """
-        Test with a single RHEL instance requesting results in futures.
+        Test with a single RHEL instance requesting results in the future.
 
-        Also assert that no ConcurrentUsage is saved to the database because the
-        requested date is in the future, and we cannot know that data yet.
+        We expect None because the we refuse to calculate data for future dates.
```
katherine-black, line 49: The we?

katherine-black (approved):

cloudigrade/api/tests/util/test_get_max_concurrent_usage.py
infinitewarp: Whoops! Good catch. :) I just amended the specific commit that introduced that change and force-pushed everything back up.

katherine-black (approved):

abellotti (approved):
took a while to go through, but yeah, LGTM !!

infinitewarp:
demo recording of the updated run and concurrent usage calculation process: https://asciinema.org/a/427887

PR #953, opened 2021-07-28
362 cloud init

katherine-black:


cloudigrade/api/clouds/aws/tasks/imageprep.py
```diff
@@ -209,8 +210,13 @@ def copy_ami_snapshot(  # noqa: C901
         create_aws_machine_image_copy(ami_id, reference_ami_id)
         ami_id = reference_ami_id
 
-    # Create volume from snapshot copy
-    create_volume.delay(ami_id, snapshot_copy_id)
+    # We only need the snapshot to launch the inspection now, so...
+    launch_inspection_instance.delay(ami_id, snapshot_copy_id)
+    # Schedule a task to cleanup the snapshot once the inspection is done.
+    delete_snapshot.apply_async(
```
infinitewarp, line 216: I assume the answer is probably "no", but do we get any indication we could process when the instance stops? Using `countdown` like this does looke like a reasonable alternative if the instance's life is a "fire and forget" situation, and this is probably fine with a large enough `INSPECTION_SNAPSHOT_CLEAN_UP_INITIAL_DELAY` value, but I wanted to ask just in case.

Is there anything involving the `DeleteOnTermination` argument or similar that could include and affect our `snapshot_copy_id`? Again, I figure this is probably "no", but I wanted to check.

infinitewarp: Even if we don't know when an instance stops, what are your thoughts on having `launch_inspection_instance` call `delete_snapshot`? If there's an unexpected backlog of `launch_inspection_instance` tasks to process, that would reduce the possibility that a long-waiting `launch_inspection_instance` doesn't have its corresponding `delete_snapshot` before it starts.

infinitewarp: Oh… maybe that risk is tempered by the `MachineImage.INSPECTED` check later in `delete_snapshot`? Still curious about your thoughts, though. 🙂

cloudigrade/api/clouds/aws/tasks/inspection.py
```diff
@@ -2,578 +2,111 @@
...
-                    "status %(status)s for instance %(ec2_instance_id)s; "
-                    "current described block device mapping is %(detail)s"
-                ),
-                {
-                    "mount_point": mount_point,
-                    "ami_id": ami_id,
-                    "ec2_instance_id": ec2_instance_id,
-                    "status": status,
-                    "detail": device_mapping,
+                "DeviceName": "/dev/xvdbb",
```
infinitewarp, line 37: Since this is always going to be a brand new instance with exactly the number of attached devices we expect, this is probably fine, but I did pause when I saw the hard-coded path/name. Just putting a note here to double-check that expectation.

cloudigrade/api/clouds/aws/tasks/inspection.py
```diff
@@ -2,578 +2,111 @@
...
-                    "value": settings.HOUNDIGRADE_SENTRY_ENVIRONMENT,
-                },
-            ]
-        )
+    houndigrade_image = (
+        f"{settings.HOUNDIGRADE_ECS_IMAGE_NAME}:{settings.HOUNDIGRADE_ECS_IMAGE_TAG}"
+    )
 
-    return container_definition
+    cloud_init_script = f"""Content-Type: multipart/mixed; boundary="//"
```
infinitewarp, line 70: Not something you need to change here now, but here's a tip that I've used elsewhere for heredocs or long strings. Since the string definition here removed the normal indentation to avoid unwanted spaces in the output, the tradeoff is a reduction in the readability and broken visual flow of the surrounding Python code. Fortunately, Python has a function to help with this kind of situation!

```python
>>> name = "potato"
>>> foo = f"""
...        hello {name}
...        this string has
...        lots of leading spaces!
...        """
>>> print(foo)

       hello potato
       this string has
       lots of leading spaces!

>>> from textwrap import dedent
>>> print(dedent(foo))

hello potato
this string has
lots of leading spaces!
```

Unfortunately, `dedent` requires _all_ not-empty lines to be indented to have the desired effect. So, in this specific code where `cloud_init_script` starts with `Content-Type` on the first line, it wouldn't work unless you moved that content down to the next line.

cloudigrade/api/clouds/aws/tasks/imageprep.py
katherine-black: So originally my plan was to put the delete call in when we processed the results, then, knowing that the AMI was inspected we would clean up the snapshot. Unfortunately we have no knowledge of what the snapshot_id is at that point, I didn't feel it was important enough to make (likely very many small) changes to pass said value around.

The snapshot is used in the RunInstance call to get it to automagically spawn and attach a volume to the instance on boot, which WILL clean up its own created volume (hence us not needing to create/delete/herd them anymore), but it will not delete the source snapshot. So that was not an option for clean up either.

So what I did is work with the following assumption, when I fire the inspection task, I know that a short time from now the image will be inspected, and the snapshot no longer valuable. So I set a delayed task to run the delete_snapshot task, which I have explicitly modified to verify that the source ami is marked as inspected. Otherwise I assume the snapshot is of value still and keep it around, as delete_snapshot will successfully exit without any deletion after creating a future task to check on it again.

cloudigrade/api/clouds/aws/tasks/inspection.py
katherine-black: Felt weird to me too, I considering defining it once at the top, but honestly this is just something I earmarked for change when/if we implement needed changes for LVM. At that point we will no longer need to designate a device name at all.

cloudigrade/api/clouds/aws/tasks/imageprep.py
katherine-black: I'll acknowledge that there is a missing case of well what if it is in error, or pending state? But I'm not sure if that has to necessarily be solved right now.

infinitewarp: This confirms my thoughts, and I'm okay with the path you've chosen. Thanks for explaining!

infinitewarp (approved):

infinitewarp (approved):

PR #965, opened 2021-08-17
Improvement for Cloud Init

katherine-black:
Two main changes here, first, we now check to make sure the snapshot is in a ready state before starting the inspection. Second, we now update the model status to inspecting at the beginning of the task.


infinitewarp (approved):
nice 👍

PR #967, opened 2021-08-17
Disable scale_up_inspection_cluster_every_60_min periodic task.

katherine-black:


infinitewarp (approved):
That `999999` isn't as big as you might think! As seconds, that's only ~11.5 days.

Regardless, the `"enabled": False` _should_ make the `schedule` value moot. 🚢 🇮🇹

infinitewarp (approved):

PR #972, opened 2021-08-23
[Draft] Accessing Sources-API via PSK Instead of x-rh-identity

abellotti:
Adding SOURCES_PSK env to clowdapp.yml
- Added new generate_sources_headers method to generate sources header with or without the PSK
- Using the new method for both Source API http calls and Kafka
- Leveraging new x-rh-sources-account-number header in inbound Kafka messages if present for fetching the account number before resorting to the x-rh-identity header.
- Updated Kafka tests
- Adding additional tests for util.redhatcloud.sources


abellotti:
Keeping as Draft for now, for this to get merged we first need for the stage and prod templates updated to define the SOURCES_PSK env for us and pulling in the new sources psk created for cloudigrade.

PR #978, opened 2021-08-27
Add pr_check and build_deploy scripts, remove tag actions flow, updat…

katherine-black:
…e push flow.

Why was the psycopg2-binary dropped in favor of building from source? Primarily security. We're standardizing on ubi8 builds on app-sre infra, we shouldn't be using any baked in libs -- https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary


infinitewarp (dismissed):

katherine-black:
> /retest

This works btw to retrigger the `PR build` job.

deployment/clowdapp.yaml
```diff
@@ -1065,6 +1106,85 @@ objects:
...
+  required: false
+  value: ''
+- name: HOUNDIGRADE_SENTRY_DSN
+  displayName: Houndigrade Sentry DSN
+  required: false
+  value: ''
+- name: DJANGO_SECRET_KEY
+  displayName: Django Secret Key
+  required: true
+  value: 'ZXBoZW1lcmFs'
```
infinitewarp, line 1126: I trust that this value isn't _actually_ a real secret since it's repeated several times below.

How would you feel about making this a more obvious placeholder value so if someone is troubleshooting they can catch it with their human eyes at a quick glance?

katherine-black: It says something like ephemeral in base64, because it feeds into a `data:` section in a secret above. Because this isn't a real template, and I can't base64 values on the fly, we have to have a base64 encoded value here.

katherine-black: also that section can't be `stringData` because otherwise ci.ext builds will have our secrets be in plaintext. Remember that output snippet I shared with you a week or so back? Yeaaa....

infinitewarp: ohhhhh gotcha! I didn't think about it as base64 when I first saw it. Thanks for the clarification!

```
echo "ZXBoZW1lcmFs" | base64 -D
ephemeral
```

infinitewarp (approved):
🚢 🇮🇹

Issue #939, opened 2021-06-30
pause and resume data collection


infinitewarp:
## Summary

As a SWatch user, I want backend data collection to pause and resume when I indicate so that I can make temporary changes to my systems without affecting the reported usage.

Origins of this issue:
- [[SPIKE] Investigate support for account disable "I don't want to delete I want to pause this for now"](https://issues.redhat.com/browse/RHCLOUD-9850)
- [#sources Slack thread started by Jacob Lindgren](https://ansible.slack.com/archives/CKJF47RS9/p1618344140119700)

cloudigrade already has internal support for this kind of functionality via `CloudAccount.disable` and `CloudAccount.enable`. We should update our Kafka message listener to process new types of message from sources-api that should trigger our code to disable and enable our account from their pause and resume updates respectively.

We should also verify with the other related services that the way we have defined our disable/enable is compatible with the way they intend to operate.

Some notes from [this Slack conversation](https://coreos.slack.com/archives/C0246P60U8H/p1625687524037700?thread_ts=1625682641.029700&cid=C0246P60U8H):
> the cliffnotes is basically:
> We broadcast an Application.pause on a pause event, and Application.unpause on unpause
> we continue to do availability checks on them, and we also disallow editing the paused resources.

Also in that conversation we discussed data gaps during the paused time period. It is okay that we collect and show no usage data during that time. This is normal and expected!

## Acceptance Criteria

- [x] Verify that "pausing" via sources in normal healthy circumstances results in us stopping data collection.
- [x] Verify that "unpausing" via sources in normal healthy circumstances results in us resuming data collection.
- [x] Verify that "unpausing" via sources when the AWS account has bad permissions, we **do not** actually enable our CloudAccount, and we notify sources with "unavailable" status. We do update our new "paused" state to False, though.
- [x] Verify that if we receive a "delete" for an application that is currently "paused", we do delete the CloudAccount like any other normal delete.

## Assumptions and Questions

- How do we handle the periodic permissions (availability check, hourly normally or half-hour when unavailable) checks during a "pause" period when our account is "disabled"?
  - cloudigrade should continue checking permissions and sending available/unavailable status updates at all times regardless of the pause state.
- What are we expected to do if we receive a "delete" while an application is "paused"?
  - cloudigrade should simply delete!
- What is the expectation on the sources side in the following:
  - source/application is healthy and running
  - customer pauses the application
  - customer breaks AWS permissions
  - sources asks cloudigrade to perform availability check
[...]

PR #981, opened 2021-08-27
support platform sources-api's new pause/resume functionality

infinitewarp:
for https://github.com/cloudigrade/cloudigrade/issues/939


cloudigrade/api/clouds/aws/tasks/verification.py
```diff
@@ -23,45 +21,23 @@ def verify_account_permissions(account_arn):
     Args:
         account_arn: The ARN for the account to check
 
-    Returns:
```
katherine-black, line 26: Should this have stayed? Since line 40 does seem to still return a bool.

infinitewarp: Good catch. I've restored the "Returns" section and rewritten the whole docstring with (hopefully) better wording in general.

katherine-black (approved):

katherine-black (approved):

PR #984, opened 2021-08-31
Add two new CG3* error_codes, wordsmith descriptions.

katherine-black:
#907

### DIFF
--- a/cloudigrade/api/error_codes.py
+++ b/cloudigrade/api/error_codes.py
@@ -166,20 +166,43 @@ CG2004 = CloudigradeError(
 # Use CG3*** for cloud account enablement issues
 CG3000 = CloudigradeError(
     "CG3000",
-    _("Failed to enable Cloud Account %(cloud_account_id)s: " "because %(exception)s"),
+    _("Failed to enable Cloud Account %(cloud_account_id)s: because %(exception)s"),
     _(
-        "Could not enable cloud metering. "
-        "Please contact support. Error code %(error_code)s."
+        "Could not enable cloud metering. Encountered unknown issue with "
+        "provided credentials. Please verify provided credentials and try "
+        "again. Error code %(error_code)s."
     ),
 )
 
 # Too many cloudtrails
 CG3001 = CloudigradeError(
     "CG3001",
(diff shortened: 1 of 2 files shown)


cloudigrade/api/error_codes.py
```diff
@@ -179,7 +180,29 @@ def notify(self, account_number, application_id, error_message=None):
     _("Failed to enable Cloud Account %(cloud_account_id)s: " "because %(exception)s"),
     _(
         "Could not enable cloud metering. You've reached the AWS CloudTrail "
-        "limit for your account. Please either increase the CloudTrail limit, or "
-        "delete an existing CloudTrail. Error code %(error_code)s."
+        "limit for your account. Please ensure you have not reached the CloudTrail "
+        "limit for the your account, and try again. Error code %(error_code)s."
```
abellotti, line 184: limit for your account

cloudigrade/api/error_codes.py
```diff
@@ -179,7 +180,29 @@ def notify(self, account_number, application_id, error_message=None):
...
+    ),
+)
+
+# Misconfigured role
+CG3002 = CloudigradeError(
+    "CG3002",
+    _("Failed to enable Cloud Account %(cloud_account_id)s: " "because %(exception)s"),
+    _(
+        "Could not enable cloud metering. An AccessDenied was received when using  "
```
abellotti, line 193: 2 space at end ?

infinitewarp, line 193: also maybe "An AccessDenied error" instead of "An AccessDenied" reads more human

cloudigrade/api/error_codes.py
```diff
@@ -168,8 +168,9 @@ def notify(self, account_number, application_id, error_message=None):
     "CG3000",
     _("Failed to enable Cloud Account %(cloud_account_id)s: " "because %(exception)s"),
```
infinitewarp, line 169: I know it wasn't part of this commit (probably an old black reformatting artifact), but there are several split strings like this in the nearby context that we could fix while we're here.
```suggestion
    _("Failed to enable Cloud Account %(cloud_account_id)s: because %(exception)s"),
```

infinitewarp (approved):

PR #987, opened 2021-09-02
Update readme and clowder playbook to use podman.

katherine-black:
Switching from docker desktop to podman should be as simple as
```
brew install podman
podman machine init
podman machine start
alias docker=podman
```

Since docker machine remains free, I did not update any of our pipelines, as that won’t be a trivial change due to questionable support for podman/buildah across github actions.

Buildah is supported as its own action, but its too orthogonal (now I’m using that word) to how we currently do image builds. With upcoming deploy changes, that may be the time to change our current docker checks to a simpler buildah bud to verify the validity of our Containerfiles.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -26,7 +26,7 @@ Running **cloudigrade** locally may require installing some or all of the follow
 -  Python 3.9
 -  `poetry <https://python-poetry.org/docs/>`_
 -  `tox <https://tox.readthedocs.io/>`_
--  `Docker Desktop <https://docs.docker.com/get-docker/>`_
+-  `Podman <https://podman.io/>`_
 -  `gettext <https://www.gnu.org/software/gettext/>`_
 -  `PostgreSQL <https://www.postgresql.org/download/>`_
 -  `AWS Command Line Interface <https://aws.amazon.com/cli/>`_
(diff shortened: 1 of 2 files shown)


infinitewarp:
I quit Docker Desktop, `brew update` and `brew upgrade`, and proceeded with the given steps:

```
brew install podman
podman machine init
podman machine start
```

…but `podman machine start` fails (or `qemu` under it fails) here:

```
INFO[0000] waiting for clients...
INFO[0000] listening tcp://0.0.0.0:7777
INFO[0000] new connection from  to /var/folders/71/5v1_8dbd03j_nfxbb8bkb1q00000gn/T/podman/qemu_podman-machine-default.sock
Waiting for VM ...
qemu-system-x86_64: warning: host doesn't support requested feature: CPUID.80000001H:ECX.svm [bit 2]
Error: dial unix /var/folders/71/5v1_8dbd03j_nfxbb8bkb1q00000gn/T/podman/podman-machine-default_ready.sock: connect: no such file or directory
```

A cursory search suggests that the "host doesn't support requested feature" error tends to appear for users with very old CPUs that don't have virtualization flags and features, but this is using a relatively modern 2020 MacBook Pro. I'll keep digging and share any more details or fixes if I find them.

infinitewarp:
Very weird. It seems that repeating `podman machine stop` and `podman machine start` after that failure _seems to_ resolve the socket file problem. It seems that the underlying `qemu` warning is still present but is not what prevented `podman` from finding/creating/using its sockets.

I'm just putting this comment here so I/we can find it in the future in case we forget.

infinitewarp:
Aside from that initial first-install problem, everything looks great! 🙂 I verified that our deploying to an ephemeral environment with our ansible playbooks continues to work too.

infinitewarp (approved):

PR #989, opened 2021-09-07
Fixed incorrect sentry dsn key usage and enablement.

abellotti:
- Added the missing accesses to the celery-sentry-dsn and listener-sentry-dsn
- Fixed the incorrect sentry dsn enablement parameters, no longer defining the
  now unused _STR entries which were remnents from the now obsolete configmap
  based parameter definitions.
- Fixed incorrect sentry environment value for the api component
- Fixed the listener's sentry parameters

### DIFF
--- a/deployment/clowdapp.yaml
+++ b/deployment/clowdapp.yaml
@@ -184,7 +184,7 @@ objects:
               key: api-sentry-dsn
               name: cloudigrade-app-secret
         - name: DJANGO_SENTRY_ENVIRONMENT
-          value: ${DJANGO_SENTRY_ENVIRONMENT}
+          value: ${API_SENTRY_ENVIRONMENT}
         - name: HOUNDIGRADE_ENABLE_SENTRY
           value: ${HOUNDIGRADE_ENABLE_SENTRY}
         - name: HOUNDIGRADE_SENTRY_DSN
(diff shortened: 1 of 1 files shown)


abellotti:
Ok, did some logging to verify we're getting the right values from the cloudigrade_app_secret keys:

- api
  - DJANGO_SENTRY_DSN   (api-sentry-dsn)
  - HOUNDIGRADE_SENTRY_DSN  (houndigrade-sentry-dsn)

- beat
  - CELERY_SENTRY_DSN (celery-sentry-dsn)

- worker
  - CELERY_SENTRY_DSN  (celery-sentry-dsn)

- listener
  - CELERY_SENTRY_DSN  (listener-sentry-dsn)

deployment/clowdapp.yaml
```diff
@@ -575,9 +578,12 @@ objects:
         - name: SENTRY_TRACES_SAMPLE_RATE
           value: ${SENTRY_TRACES_SAMPLE_RATE}
         - name: CELERY_ENABLE_SENTRY
-          value: ${CELERY_ENABLE_SENTRY}
+          value: ${LISTENER_ENABLE_SENTRY}
         - name: CELERY_SENTRY_DSN
```
katherine-black, line 582: `CELERY_*` sentry env vars are only used in the celery.py config file, so they'll only ever get used by the beat and worker. As the listener is a django management command, you'll want to populate the `DJANGO_*` sentry env vars that are used in prod.py with the listener DSN and other values.

katherine-black (changes requested):

abellotti:
@kat commit: https://github.com/cloudigrade/cloudigrade/pull/989/commits/2f1a71737848816d00c9fd4de6e51d34821576df pushed. I `think` that takes care of it. Let me know if I missed anything else.  Thanks.

deployment/clowdapp.yaml
abellotti: @kat commit: https://github.com/cloudigrade/cloudigrade/pull/989/commits/2f1a71737848816d00c9fd4de6e51d34821576df pushed. I `think` that takes care of it. Let me know if I missed anything else.  Thanks.

katherine-black (approved):
Looks good, thank you!

PR #996, opened 2021-09-15
Switch to using sources-monitor-go

katherine-black:
sources-monitor appears to have become sources-monitor-go, this updates our clowdapp and playbook to reference that accordingly.


abellotti (approved):

abellotti:
👍 tripped over that yesterday.

PR #1001, opened 2021-09-20
Bumping up the CELERY_WORKER_LIMITS_CONTAINER_MEM_LIMIT

abellotti:
- Bumping up the CELERY_WORKER_LIMITS_CONTAINER_MEM_LIMIT by 384Mi to cover the memory spikes of 316Mi seen
with the Azure CLI SDK. So going from 512Mi to 896Mi.


abellotti:
Keeping as draft for now until the dust settles with our clowder deployments.

PR #1006, opened 2021-09-20
Have ansible base64 encode env vars before passing them into secrets.

katherine-black:


abellotti:
When an env value is blank, do we want the base64 Cg== for blank or just "" in the secret ?

katherine-black:
@abellotti We want the base64 value, as it's being fed into a `data` field that expects only base64 values.

katherine-black:
Also, technically, one should be providing all of those values. Hypothetically "" could be accepted, but I don't see a need to allow for that as a blank value, base64 or not, will lead to things not working. So in the end it matters not :)

infinitewarp (approved):

PR #1064, opened 2021-11-04
Downgraded the MachineImage.DoesNotExist error saving an instance

abellotti:
- Downgraded the MachineImage.DoesNotExist error while trying to save an instance to just a logger info.
- Updated the test to check the condition was logged.

https://github.com/cloudigrade/cloudigrade/issues/1055

### DIFF
--- a/cloudigrade/api/clouds/aws/util.py
+++ b/cloudigrade/api/clouds/aws/util.py
@@ -342,8 +342,7 @@ def save_instance(account, instance_data, region):
             logger.info(
                 _(
                     "Existing AwsMachineImage %(awsmachineimage_id)s "
-                    "(ec2_ami_id=%(ec2_ami_id)s) found has no "
-                    "MachineImage. This should not happen!"
+                    "(ec2_ami_id=%(ec2_ami_id)s) found has no MachineImage."
                 ),
                 {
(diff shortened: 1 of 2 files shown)


infinitewarp (approved):
👍

cloudigrade/api/clouds/aws/util.py
```diff
@@ -339,7 +339,7 @@ def save_instance(account, instance_data, region):
             # We are not sure how this could happen. Whenever we save a new
             # AwsMachineImage, we *should* always follow up with creating
             # its paired MachineImage. Investigate if you see this error!
-            logger.error(
+            logger.info(
                 _(
                     "Existing AwsMachineImage %(awsmachineimage_id)s "
                     "(ec2_ami_id=%(ec2_ami_id)s) found has no "
```
katherine-black, line 345: We should drop the `This should not happen!` from the line below, as this happens all the time :)

katherine-black (changes requested):

katherine-black (approved):

cloudigrade/api/clouds/aws/util.py
katherine-black: Thank you!

PR #1076, opened 2021-11-16
Adding support for SonarQube in PR Checks.

abellotti:
- Adding support for SonarQube in PR Checks.

Requires: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/28981

### DIFF
--- a/deployment/scripts/sonarqube.sh
+++ b/deployment/scripts/sonarqube.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+
+echo "SonarQube Scan"
+echo "Directory: $PWD"
+echo "JAVA:      $JAVA_HOME"
+
+mkdir $PWD/sonarqube/
+mkdir $PWD/sonarqube/download/
+mkdir $PWD/sonarqube/extract/
+mkdir $PWD/sonarqube/certs/
+mkdir $PWD/sonarqube/store/
+
+curl -o $PWD/sonarqube/certs/RH-IT-Root-CA.crt --insecure $RH_IT_ROOT_CA_CERT_URL
+
+$JAVA_HOME/bin/keytool \
+  -keystore /$PWD/sonarqube/store/RH-IT-Root-CA.keystore \
+  -import \
+  -alias RH-IT-Root-CA \
+  -file /$PWD/sonarqube/certs/RH-IT-Root-CA.crt \
+  -storepass redhat \
+  -noprompt
+
+export SONAR_SCANNER_OPTS="-Djavax.net.ssl.trustStore=$PWD/sonarqube/store/RH-IT-Root-CA.keystore -Djavax.net.ssl.trustStorePassword=redhat"
+
+
+export SONAR_SCANNER_OS="linux"
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    export SONAR_SCANNER_OS="macosx"
+fi
+
+export SONAR_SCANNER_CLI_VERSION="4.6.2.2472"
+export SONAR_SCANNER_DOWNLOAD_NAME="sonar-scanner-cli-$SONAR_SCANNER_CLI_VERSION-$SONAR_SCANNER_OS"
+export SONAR_SCANNER_NAME="sonar-scanner-$SONAR_SCANNER_CLI_VERSION-$SONAR_SCANNER_OS"
+
+curl -o $PWD/sonarqube/download/$SONAR_SCANNER_DOWNLOAD_NAME.zip --insecure https://binaries.sonarsource.com/Distribution/sonar-scanner-cli/$SONAR_SCANNER_DOWNLOAD_NAME.zip
+
+unzip -d $PWD/sonarqube/extract/ $PWD/sonarqube/download/$SONAR_SCANNER_DOWNLOAD_NAME.zip
+
+export PATH="$PWD/sonarqube/extract/$SONAR_SCANNER_NAME/bin:$PATH"
+
+COMMIT_SHORT=$(git rev-parse --short=7 HEAD)
+
+sonar-scanner \
+  -Dsonar.projectKey=console.redhat.com:cloudigrade \
+  -Dsonar.sources=./cloudigrade \
+  -Dsonar.host.url=https://sonarqube.corp.redhat.com \
+  -Dsonar.projectVersion=$COMMIT_SHORT \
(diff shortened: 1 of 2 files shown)


deployment/scripts/sonarqube.sh
```diff
@@ -0,0 +1,55 @@
...
+export SONAR_SCANNER_OS="linux"
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    export SONAR_SCANNER_OS="macosx"
+fi
+
+export SONAR_SCANNER_CLI_VERSION="4.6.2.2472"
+export SONAR_SCANNER_DOWNLOAD_NAME="sonar-scanner-cli-$SONAR_SCANNER_CLI_VERSION-$SONAR_SCANNER_OS"
+export SONAR_SCANNER_NAME="sonar-scanner-$SONAR_SCANNER_CLI_VERSION-$SONAR_SCANNER_OS"
+
+curl -o $PWD/sonarqube/download/$SONAR_SCANNER_DOWNLOAD_NAME.zip --insecure https://binaries.sonarsource.com/Distribution/sonar-scanner-cli/$SONAR_SCANNER_DOWNLOAD_NAME.zip
```
katherine-black, line 35: Maybe left over `--insecure`? We're downloading a binary file from a foreign domain, we should verify the certificate is valid 🙂

abellotti: 👍

deployment/scripts/sonarqube.sh
```diff
@@ -0,0 +1,55 @@
...
+echo "Directory: $PWD"
+echo "JAVA_HOME: $JAVA_HOME"
+
+mkdir $PWD/sonarqube/
+mkdir $PWD/sonarqube/download/
+mkdir $PWD/sonarqube/extract/
+mkdir $PWD/sonarqube/certs/
+mkdir $PWD/sonarqube/store/
+
+curl -o $PWD/sonarqube/certs/RH-IT-Root-CA.crt --insecure $RH_IT_ROOT_CA_CERT_URL
```
infinitewarp, line 13: Is this `--insecure` necessary?

deployment/scripts/sonarqube.sh
```diff
@@ -0,0 +1,55 @@
...
+
+sonar-scanner \
+  -Dsonar.projectKey=console.redhat.com:cloudigrade \
+  -Dsonar.sources=./cloudigrade \
+  -Dsonar.host.url=https://sonarqube.corp.redhat.com \
+  -Dsonar.projectVersion=$COMMIT_SHORT \
+  -Dsonar.login=$SONARQUBE_TOKEN
+
+mkdir -p $WORKSPACE/artifacts
+cat << EOF > ${WORKSPACE}/artifacts/junit-dummy.xml
```
infinitewarp, line 51: What is the purpose of this dummy file?

abellotti: not quite sure, copied from the documented PRs in the SonarQube google doc,  /cc @chambridge would you know offhand ?

chambridge: pr-check job always checks for a junit artifact. If its not present the job will fail, this essentially just side-steps that by creating an artifact.

deployment/scripts/sonarqube.sh
abellotti: I don't think so, Just tried both ways. I'll remove it.

deployment/scripts/sonarqube.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
+echo "SonarQube Scan"
+echo "Directory: $PWD"
```
infinitewarp, line 4: What is `$PWD` expected to return at this point? Do we know the runner's current working directory on the filesystem when this script is called?

I ask because I'd be concerned that its value might change unexpectedly if the runner itself is ever called differently or if the runner ever calls this script differently than usual. Since we're `mkdir`ing a bunch of extra folders and files, that could put things in different places, and I don't know if that's okay. 

Should use something more predictable and static? How about something `/tmp/` or `$TMPDIR`? Or something like [this $SCRIPT_DIR one-liner](https://stackoverflow.com/a/246128)?

deployment/scripts/sonarqube.sh
```diff
@@ -0,0 +1,55 @@
...
+mkdir $PWD/sonarqube/
+mkdir $PWD/sonarqube/download/
+mkdir $PWD/sonarqube/extract/
+mkdir $PWD/sonarqube/certs/
+mkdir $PWD/sonarqube/store/
+
+curl -o $PWD/sonarqube/certs/RH-IT-Root-CA.crt --insecure $RH_IT_ROOT_CA_CERT_URL
+
+$JAVA_HOME/bin/keytool \
+  -keystore /$PWD/sonarqube/store/RH-IT-Root-CA.keystore \
```
infinitewarp, line 16: For consistency with the other commands, especially if `$PWD` ever contains a relative path, these should drop the leading `/`s.

deployment/scripts/sonarqube.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
+echo "SonarQube Scan"
+echo "Directory: $PWD"
+echo "JAVA_HOME: $JAVA_HOME"
+
+mkdir $PWD/sonarqube/
```
infinitewarp, line 7: Here and throughout the script, any variables like `$PWD` that define filesystem paths need to be quoted like `"$PWD"` (or ideally like `"${PWD}"` also to prevent potential variable resolution issues).

Variables in paths _should always_ be quoted to minimize the risk of unexpected characters blowing up your commands. For example, if any directory in the PWD stack contains a space in its name, if you don't quote the variable, the space character will break the one command argument into two parts, and bad things could happen.

deployment/scripts/sonarqube.sh
abellotti: That value seems to be /var/lib/jenkins/workspace/cloudigrade-cloudigrade-pr-check, so probably a good idea to keep  things as Jenkins probably clean things up afterward.

deployment/scripts/sonarqube.sh
infinitewarp: Thanks @chambridge ! Would you throw a brief comment in the script with that explanation, @abellotti ?

infinitewarp (approved):
👍

PR #1082, opened 2021-11-30
Add image tag to slack payload

katherine-black:
Add IMAGE_TAG to slack payload, so we can quickly see if we're having new or old rollouts in stage/prod.


infinitewarp (approved):
Good idea!

Issue #1063, opened 2021-11-01
More dedicated queues for Celery tasks


infinitewarp:
## Summary

As a cloudigrade sysop, I want to more clearly understand which tasks are piling up in the queues when we see a large number so that I can better diagnose and resolve the misbehaving task.

We recently experienced an incident that spawned 100,000+ messages on the default Celery queue. Unfortunately, there are several tasks that use the default queue, and this made troubleshooting difficult. We should update our configuration so that all current Celery tasks use their own respective queue.

## Acceptance Criteria

- [x] Verify via code review that all tasks have configured queues
- [x] Verify that cloudigrade still functions normally
- [x] Verify via an actual deployment that the new queues exist in AWS SQS

## Assumptions and Questions

- When we deploy this to stage and prod, we should check the old Celery queue for any unread messages and requeue them to their new queues. How do we do this? One-off manual script run?
- The original issue that caused 100,000+ messages to be in queue has been resolved independently and is part of this issue. That was caused by our deployment having too few workers and being unable to process enough messages quickly. That was resolved by adding more workers.

abellotti:
**Cloudigrade Celery Tasks:**

____

- api/clouds/aws/tasks/__init__.py

| Task | In Route | In Worker |
| ---- | :------: | :-------: |
| api.clouds.aws.tasks.attach_volumes_to_cluster                       | n | n |
| api.clouds.aws.tasks.create_volume                                   | n | n |
| api.clouds.aws.tasks.enqueue_ready_volume                            | n | n |
| api.clouds.aws.tasks.run_inspection_cluster                          | n | n |
| api.clouds.aws.tasks.scale_down_cluster                              | n | n |
| api.clouds.aws.tasks.scale_up_inspection_cluster                     | n | n |
| api.clouds.aws.tasks.verify_verify_tasks                             | n | n |

____

- api/clouds/aws/tasks/cloudtrail.py

| Task | In Route | In Worker |
| ---- | :------: | :-------: |
| api.clouds.aws.tasks.analyze_log                                     | y | n |

____

- api/clouds/aws/tasks/imageprep.py

| Task | In Route | In Worker |
| ---- | :------: | :-------: |
| api.clouds.aws.tasks.copy_ami_snapshot                               | y | n |
| api.clouds.aws.tasks.copy_ami_to_customer_account                    | y | n |
| api.clouds.aws.tasks.delete_snapshot                                 | y | n |
[...]

PR #1085, opened 2021-12-01
Assure that all celery tasks are properly defined

abellotti:
- Update to assure that all celery tasks are properly declared in the celery route as well as in the list of queues in the worker.
https://github.com/cloudigrade/cloudigrade/issues/1063

### DIFF
--- a/cloudigrade/config/settings/base.py
+++ b/cloudigrade/config/settings/base.py
@@ -451,14 +451,47 @@ SCHEDULE_VERIFY_VERIFY_TASKS_INTERVAL = env.int(
 
 CELERY_TASK_ROUTES = {
     # api.tasks
-    "api.tasks.delete_cloud_account": {"queue": "delete_cloud_account"},
-    "api.tasks.delete_inactive_users": {"queue": "delete_inactive_users"},
-    "api.tasks.inspect_pending_images": {"queue": "inspect_pending_images"},
-    "api.tasks.persist_inspection_cluster_results_task": {
-        "queue": "persist_inspection_cluster_results_task"
+    "api.tasks.calculate_max_concurrent_usage": {
+        "queue": "calculate_max_concurrent_usage"
+    },
+    "api.tasks.delete_cloud_account": {
+        "queue": "delete_cloud_account"
+    },
+    "api.tasks.enable_account": {
+        "queue": "enable_account"
+    },
+    "api.tasks.delete_inactive_users": {
+        "queue": "delete_inactive_users"
     },
     "api.tasks.delete_orphaned_cloud_accounts": {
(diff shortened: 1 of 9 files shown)


cloudigrade/config/settings/base.py
```diff
@@ -448,15 +448,48 @@ def __print(*args, **kwargs):
...
-    "api.tasks.delete_inactive_users": {"queue": "delete_inactive_users"},
-    "api.tasks.inspect_pending_images": {"queue": "inspect_pending_images"},
-    "api.tasks.persist_inspection_cluster_results_task": {
-        "queue": "persist_inspection_cluster_results_task"
+    "api.tasks.calculate_max_concurrent_usage": {
+        "queue": "calculate_max_concurrent_usage"
+    },
+    "api.tasks.delete_cloud_account": {
+        "queue": "delete_cloud_account"
+    },
```
infinitewarp, line 456: Do you have your IDE or other local tooling configured to run `black` automatically against your code?

On my system, `black` removes excess whitespace and collapses lines like this, and the result looks like the previous version of the file like this:

```
    "api.tasks.delete_cloud_account": {"queue": "delete_cloud_account"},
```

I wonder if there's something odd about our `tox` config that isn't catching this. 🤔

infinitewarp: Well, yep, I just discovered that we might be a bit too aggressive with our ignore pattern in `tox.init`. I may address this in a separate PR.

infinitewarp: My comments will be resolved soon with https://github.com/cloudigrade/cloudigrade/pull/1087

cloudigrade/config/settings/base.py
```diff
@@ -467,47 +490,57 @@ def __print(*args, **kwargs):
     "api.tasks.delete_from_sources_kafka_message": {
         "queue": "delete_from_sources_kafka_message"
     },
-    "api.tasks.update_from_sources_kafka_message": {
-        "queue": "update_from_sources_kafka_message"
+    "api.tasks.update_from_source_kafka_message": {
+        "queue": "update_from_source_kafka_message"
```
infinitewarp, line 494: Interesting. 🤔 How would you feel about instead renaming the underlying function to have plural `sources` instead of the singular `source`? That would make it more consistent with the other names, and that's probably what we intended before this bug.

abellotti: I don't mind updating in this PR, consistency is always good. 👍

infinitewarp (approved):
LGTM!

abellotti:
Thanks @infinitewarp for the 👀  and review.

PR #1086, opened 2021-12-02
update to ubi 8.5

infinitewarp:


katherine-black (changes requested):
We should also bump https://github.com/cloudigrade/cloudigrade/blob/master/deployment/clowdapp.yaml#L1623 when we have a new houndigrade artifact built on ubi8.5

katherine-black (approved):

PR #1087, opened 2021-12-02
stricter tox/flake8 ignore and exclude rules

infinitewarp:


tox.ini
```diff
@@ -15,11 +15,13 @@ ignore = D106,D212,W503,W504,E203
 extend-ignore = E231
 max-complexity = 10
 exclude =
-  **/__init__.py
-  **/manage.py
```
katherine-black, line 19: Why lint a framework provided file?

infinitewarp: Arguably, once it's generated by the framework, it's ours to maintain. Besides, it was trivial to get it passing checks.

katherine-black (approved):

Issue #905, opened 2021-05-14
cloud.redhat.com onboarding phase 2


infinitewarp:
## Summary

As a cloudigrade sysop, I want my service "fully onboarded" with SRE so that they can carry the pager and we can get better support in stage and prod.

This process includes but may not be limited to:

- ~[Review ACs for CloudDot Cloudigrade](https://issues.redhat.com/browse/APPSRE-3832)~ closed 2022-07-11
- ~[APPSRE Onboarding: Implement ACs for CloudDot Cloudigrade](https://issues.redhat.com/browse/CLOUDMETER-2)~ closed 2022-04-19
- ~[APPSRE Onboarding: Implement SLIs and SLOs for CloudDot Cloudigrade](https://issues.redhat.com/browse/CLOUDMETER-3)~ closed 2022-04-18
- ~[[cloudmeter] fill the AppSRE onboarding questionnaire](https://issues.redhat.com/browse/RHCLOUD-14186)~ closed 2021-10-31
- ~Google Docs: [Cloudmeter | AppSRE Onboarding Questionnaire](https://docs.google.com/document/d/1GFn5An1rDZ18Wln3kCd5kDQ9T6ZjLZlpflJoBhYsZ34/edit?usp=sharing)~ old
- ~Google Sheets: [Cloud meter Acceptance Criteria Checklist](https://docs.google.com/spreadsheets/d/1mgiQA2dG4bnijCTlaBmKC5Ns7-AjMiPVhStQQokWtz0/edit#gid=0)~ "accepted" 2022-07-11
- [APPSRE-3836: SRE Checkpoint for CloudDot Cloudigrade](https://issues.redhat.com/browse/APPSRE-3836)

Key checklist issues we need to resolve include:

- define an Architecture Document
  - consider using [this example](https://osbuild.pages.redhat.com/internal-guides/image-builder-service/architecture/image-builder-crc.html) as a template
  - "The service owner must document the expected growth of the resource usage, including CPU, Memory, Network and Storage in the AppSRE Onboarding Questionnaire."
  - "Identify where service data resides. Document the impact of losing said data in the architecture document."
  - "The service contains an Architecture Document which includes the information specified in this section."
  - "The technology stack is documented in the architecture document."
- "Any ad-hoc steps required to recover the service upon a disaster event have been documented in an SOP."
  - [example reference MR](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/25720)
- SLIs
[...]

PR #1091, opened 2021-12-13
update Grafana dashboard better to suit upcoming SRE alerts and SLIs

infinitewarp:
Custom Grafana dashboards are periodically deleted in stage, but for now a live copy exists [here](https://grafana.stage.devshift.net/d/null/brasmith-testing).

Notable changes:
- Display the current pod count and ratio of ready pods to the replica count.
- Update how CPU and memory usage are calculated, hopefully more correctly now.
- Add counts of 3xx responses (previously only 2xx, 4xx, and 5xx).
- Show rates of all responses by HTTP status.
- Show average response times of all responses by HTTP status.
- Use color to indicate abnormal values in some fields. For example, green if 1 ready beat pod but red if more than 1.

For https://github.com/cloudigrade/cloudigrade/issues/905

### DIFF
--- a/dashboards/grafana-dashboard-clouddot-insights-cloudigrade.configmap.yaml
+++ b/dashboards/grafana-dashboard-clouddot-insights-cloudigrade.configmap.yaml
@@ -11,16 +11,23 @@ data:
             "hide": true,
             "iconColor": "rgba(0, 211, 255, 1)",
             "name": "Annotations & Alerts",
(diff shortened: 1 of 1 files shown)


infinitewarp:
Let me know if the stage dashboard disappears before you get to review it, and I'll create another.

katherine-black (approved):

Issue #1089, opened 2021-12-08
Drop `AwsCloudAccount.verify_task` and related scaffolding


infinitewarp:
## Summary

As a cloudigrade developer and sysop, I want our account verification process to be only as complex as necessary so that cloudigrade is easier to maintain. I also want to rely on sources-api to trigger account verification.

We currently create a `PeriodicTask` associated with every `AwsCloudAccount` that ultimately calls `api.clouds.aws.tasks.verify_account_permissions` every `SCHEDULE_VERIFY_VERIFY_TASKS_INTERVAL` seconds (24 hours by default). This is normally redundant, though, because sources-api already calls our `availability_check` endpoint every 60 minutes for an enabled source or every 30 minutes for a disabled source. We have been assured that this process is reliable and that things must be going very badly if it ever stops.

So, we should reduce our complexity by removing `AwsCloudAccount.verify_task` and all the related scaffolding for it, including but not necessarily limited to:

- `AwsCloudAccount._enable_verify_task`
- `AwsCloudAccount._disable_verify_task`
- `api.clouds.aws.tasks.verify_account_permissions` is the task function being called
- `verify_verify_tasks` which was deprecated long ago, and should be safe to finally remove now.
- `api.clouds.aws.tasks.ensure_all_verify_tasks_are_valid` which checks existence of tasks for accounts

## Acceptance Criteria

- [x] Verify the above mentioned functions are gone via code review
- [x] Verify cloudigrade continues functioning normally (regression testing)

## Assumptions and Questions

- If during development you find there are unexpected inter-dependencies that may make this more challenging than expected or that reveal an edge case that we didn't realize we still need, please reach out to the team to discuss.
- We expect that Azure will follow the same type of process as we have for AWS. soruces-api will poll our availability check for any Azure sources, and we will rely on that instead of keeping our own scheduled tasks. We believe that we currently have no scheduled tasks for Azure, but if any exist, they should be deleted like the AWS ones.

infinitewarp:
This is done, and everything still seems stable. Closing!

PR #1096, opened 2021-12-20
remove CloudAccount.verify_task and related functionality

infinitewarp:
https://github.com/cloudigrade/cloudigrade/issues/1089

also require Celery ~= 5.2.1 to ensure we always get Kombu 5.2.0+


deployment/clowdapp.yaml
```diff
@@ -1497,9 +1488,7 @@ parameters:
     scale_down_cluster,\
     scale_up_inspection_cluster,\
```
abellotti, line 1489: looks like some of those other tasks that were removed also need to be removed from the worker here.

katherine-black (approved):

deployment/clowdapp.yaml
infinitewarp: Good catch! I've fixed this in a subsequent push, and I've also added a script to our CI to make sure we never miss something like this again (hopefully).

infinitewarp:
I added a new command to sanity-check our various Celery configs. I also deliberately let a recent build fail so I could see the result in the PR here.

https://github.com/cloudigrade/cloudigrade/runs/4612193049?check_suite_focus=true

Screen Shot 2021-12-22 at 5 20 22 PM

Screen Shot 2021-12-22 at 5 19 07 PM

This should keep us from forgetting related changes going forward.

katherine-black (approved):

PR #1099, opened 2021-12-22
Added task to delete cloud accounts not in sources

abellotti:
- Added the periodic task delete_cloud_accounts_not_in_sources to go through the database and delete inactive accounts with the related Aws/Azure content objects that do not exit in sources.
- Also adding the /internal/delete_cloud_accounts_not_in_sources endpoint to allows us to trigger it outside its schedule.


cloudigrade/config/settings/prod.py
```diff
@@ -86,7 +86,11 @@ def django_traces_sampler(sampling_context):
         send_default_pii=True,
     )
 
-# Specifically in production, this default should be False.
+# Specifically in production, these defaults should be False.
```
infinitewarp, line 89: Shouldn't this be the other way around? We _do_ want this enabled in prod, right?

I know this gets set/overridden later in the `clowdapp.yaml`, but maybe it's finally time for this (and the `SOURCES_ENABLE_DATA_MANAGEMENT_FROM_KAFKA` setting below it) to have a more reasonable default.

Thoughts?

infinitewarp:
This generally looks good to me! 🙂 I just have that one configuration defaults question.

cloudigrade/config/settings/prod.py
katherine-black: I'll +1 @infinitewarp, I think this functionality should be enabled by default in all environments, it seems fundamental to cloudigrade not retaining stale data.

abellotti: makes sense  👍   I'll update both to default to True.

katherine-black (approved):

infinitewarp (approved):
Looks good!

PR #1113, opened 2022-01-26
Wrap 'launch_inspection_instance' task in a transaction.

katherine-black:
#1092 
launch_inspection_instance was missing a @transaction.atomic decorator.


infinitewarp (approved):
Good catch. Thanks!

infinitewarp (approved):
👌

PR #1114, opened 2022-01-27
log more verbosely when we fail to read/parse messages from SQS and S3

infinitewarp:


katherine-black (changes requested):
Looks simple and straightforward, lets just get those tests in and we can get this out 👍

katherine-black (approved):
👍

Issue #661, opened 2021-01-28
Flower for Celery


katherine-black:
In GitLab by @infinitewarp on Mar 24, 2020, 17:52

## Summary

As a cloudigrade sysop, I would like to have better visibility into what Celery tasks are doing. [Flower](https://flower.readthedocs.io/en/latest/) is a web based tool for monitoring and administrating Celery clusters.

The overall "ask" of this is only to deploy Flower so we can see and interact with its controls. Additional features like Prometheus integration will come later.

## Acceptance Criteria

- [ ] Verify a Flower deployment exists with each of our other clowder deployments
  - [ ] Ephemeral
  - [ ] Stage
  - [ ] Prod
- [ ] Verify routes are available so that we can see the web interface.
  - [ ] Internal only, not part of the public API.
  - [ ] Stage and prod web interfaces are made available via Turnpike using the same doppler-dev group.
- [ ] Verify that the web interface shows accurate and meaningful data.

## Assumptions and Questions

- Our deployment will be like beat and worker in that it will be part of the cloudigrade app, not deployed separately like postigrade.
- QE does not need to build integration tests for this issue.
- koku did this and [may be a reference implementation](https://github.com/RedHatInsights/e2e-deploy/blob/master/templates/hccm-optional/celery-flower.yaml).

infinitewarp:
Blocked until we do https://github.com/cloudigrade/cloudigrade/issues/433 because Flower uses [RabbitMQ Managment Plugin](https://www.rabbitmq.com/management.html) to get info about queues. ([according to the current docs](https://flower.readthedocs.io/en/latest/config.html#broker-api)).

infinitewarp:
https://github.com/cloudigrade/cloudigrade/pull/1105 has merged, and our Celery is now using Redis instead of SQS! 🎉 

I replaced the `blocked` label with `needs-definition` here so we don't forget to discuss this important next step. Also, the RabbitMQ Management Plugin _is_ still referenced by the Flower docs, which is mildly concerning, but I think that's only relevant _if_ you're using RabbitMQ as your broker.

abellotti:
- Chatted a bit with Chris H. yesterday, apparenly Koku no longer uses flower. It was never moved to OCP4.  Koku was only using flower on OCP3 and was using the RabbitMQ broker. They decided to drop flower in OCP4 in favor of using a couple of specific turnpike APIs and moving to Redis. They also did not want to continue managing credential issues and associated CVE updates.
[...]

PR #1117, opened 2022-02-07
Adding deployment of flower for Celery

abellotti:
- Added flower dependency via poetry
- Updated clowdapp.yaml to create a single cloudigrade-flower pod similar to beat and start flower on port 8000.
- Created the flowerconfig.py to define some of the flower configs that are based on settings.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,6 +32,7 @@ RUN microdnf install -y \
     && pip install -U pip \
     && pip install poetry \
     && poetry config virtualenvs.in-project true \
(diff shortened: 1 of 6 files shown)


abellotti:
Keeping this as Draft for now, as the latest published flower (1.0.0), does not work with Celery. Will need a published 1.0.1 flower to PyPi. See https://github.com/cloudigrade/cloudigrade/issues/661 for more details.  Bumping down flower to 0.9.7 and Celery to to 4.4.7, allowed us to fetched all the goodness from the redis broker.

Issue #1118, opened 2022-02-09
Spike: Investigate methods of Redis/Celery-to-Prometheus


infinitewarp:
## Summary

As a cloudigrade sysop, I want monitoring and alerting for Redis availability and queue sizes so that I can react to new problems as they arise.

How do we get the useful data into Prometheus? 

Options to consider and research on this issue:

- [celery-exporter](https://github.com/danihodovic/celery-exporter)?
  - @infinitewarp found discussion in StackOverflow suggesting its use or maybe a fork of it?
  - need to verify which fork is active and maintained
  - need to verify if it actually works with Celery/Kombu/etc. latest versions.
- manually sending commands to Redis?
  - where would we do this? part of our health check?
  - would we expose the values we poll into the existing Prometheus metrics exporter endpoint?
  - via Python libraries? If so, which?
  - or via execing out shell commands to `redis-cli`? hopefully not, but a last resort.
- other ideas?

@abellotti is also asking around to see if/how Koku and friends are handling this.

Expected output: Answers to the above questions with some basic prototype experiments.

Time box: 1 week

abellotti:
It's a beautiful thing:

```
$ poetry run manage.py shell
...
Python 3.10.1 (main, Dec  6 2021, 22:25:40) [Clang 13.0.0 (clang-1300.0.29.3)]
Type 'copyright', 'credits' or 'license' for more information
IPython 7.31.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from config.celery import app

In [2]: conn = app.connection()

In [3]: inspector = app.control.inspect(connection=conn, timeout=10)

In [4]: print(inspector.active())
[...]

PR #1128, opened 2022-02-28
Adding support for Celery Metrics

abellotti:
- Based on celery-exporter
- Exposed as part of /internal/metrics
- Celery metrics listener in a separate thread

For https://github.com/cloudigrade/cloudigrade/issues/1118

### Metrics
Name     | Description | Type
---------|-------------|----
celery_task_sent_total | Sent when a task message is published. | Counter
celery_task_received_total | Sent when the worker receives a task. | Counter
celery_task_started_total | Sent just before the worker executes the task. | Counter
celery_task_succeeded_total | Sent if the task executed successfully. | Counter
celery_task_failed_total | Sent if the execution of the task failed. | Counter
celery_task_rejected_total | The task was rejected by the worker, possibly to be re-queued or moved to a dead letter queue. | Counter
celery_task_revoked_total | Sent if the task has been revoked. | Counter
celery_task_retried_total | Sent if the task failed, but will be retried in the future. | Counter
celery_worker_up | Indicates if a worker has recently sent a heartbeat. | Gauge
celery_worker_tasks_active | The number of tasks the worker is currently processing | Gauge
celery_task_runtime | Histogram of runtime measurements for each task | Histogram

### DIFF
--- a/cloudigrade/internal/metrics.py
+++ b/cloudigrade/internal/metrics.py
@@ -88,25 +88,31 @@ class CeleryMetrics:
             _("Received celery event='%s' for task='%s'"), event["type"], task.name
         )
 
-        counter = self.state_counters.get(event["type"])
-        if not counter:
+        if event["type"] not in self.state_counters:
             logger.warning(_("No celery counter matches task state='%s'"), task.state)
-            return
 
-        labels = {}
-        # pylint: disable=protected-access
-        for labelname in counter._labelnames:
-            value = getattr(task, labelname)
-            if labelname == "exception":
-                logger.debug(value)
-                value = get_exception_class(value)
-            labels[labelname] = value
+        labels = {"name": task.name, "hostname": task.hostname}
 
-        counter.labels(**labels).inc()
-        logger.debug(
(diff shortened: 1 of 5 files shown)


cloudigrade/internal/metrics.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+class CeleryMetrics:
+    """Celery Metrics class for processing and exposing Prometheus metrics."""
+
+    def __init__(self, interval=2, buckets=None):
+        """Celery Metrics Initializer."""
+        self.listener_started = False
+        self.interval = interval
+        self.state_counters = {
```
infinitewarp, line 23: I'm not familiar with these names, aside from the obvious inline strings that describe them. Where did you find them to include here?

infinitewarp (changes requested):

cloudigrade/internal/metrics.py
abellotti: borrowed/adapted from celery-exporter - https://github.com/danihodovic/celery-exporter/blob/master/src/exporter.py

PR #1133, opened 2022-03-09
More refactoring cleanup

infinitewarp:
Here are some more slight changes and code reorgs that I've split from other local work. I'm trying to get these out ahead of the rest so there isn't just one massive PR to review.

Here's a breakdown of the changes. Also, each commit in this PR is atomic with an appropriate message.

- Very slightly reorganized `internal/urls.py` and added some inline comments to improve readability.
- Moved `/internal/api/cloudigrade/v1/sources_kafka/` to `/internal/sources_kafka/` because the rest of `/internal/api/cloudigrade/v1/` is more an internal reflection of our public API and otherwise only contains "simple" model viewsets. This was originally going to be part of a more aggressive internal API path reorganization, but I backed off that idea for now since sources and tally may be using some of the internal APIs, and I don't want to deal with a deprecation/migration plan right now.
- Moved the standard Django admin out of `/internal/` and added conditions to remove it from production deployments. This is a bit of a subjective change, and I'd be open to feedback, but I though it made sense to live outside of `/internal/` since *everything else* under `/internal/` are _our_ custom endpoints. I was also tempted to just drop the admin entirely since we've never implemented anything in there.
- Split and moved the custom ViewSets classes from function-based views into appropriately-named packages. In my _other_ branch, I have been adding new internal view functions and classes, and I noticed the internal one-package-to-rule-them-all was getting a bit long.


cloudigrade/config/settings/prod.py
```diff
@@ -8,6 +8,10 @@
 
 IS_PRODUCTION = CLOUDIGRADE_ENVIRONMENT == "prod"
 
+if IS_PRODUCTION and "django.contrib.admin" in DJANGO_APPS:
+    # Do not allow the standard Django admin in production environments.
+    DJANGO_APPS.remove("django.contrib.admin")
```
katherine-black, line 13: okaywiththis.jpg

cloudigrade/config/urls.py
```diff
@@ -1,7 +1,17 @@
...
     path("internal/", include("internal.urls")),
 ]
+
+if not settings.IS_PRODUCTION:
+    # Allow the standard Django admin only in not-production environments.
+    from django.contrib import admin
+
+    urlpatterns += [
+        path("api-auth/", include("rest_framework.urls")),
+        path("admin/", admin.site.urls),
```
katherine-black, line 16: I strongly approve of this change.

katherine-black (approved):
Thank you for cleaning up those paths, great changes!

PR #1139, opened 2022-03-22
Create new "synthetic data request" API to replace `spawndata` management command

infinitewarp:
The gist of this change is to create a new internal HTTP API that we can use to synthesize pseudorandom customer activity data for testing in various environments. Behaviorally, this is very similar to the `spawndata` management command; it's inputs and outputs are very similar (though not exactly the same in implementation). We cannot easily use `spawndata` in most non-local deployed environments, though, because it requires direct shell access to the system. Also, `spawndata` can take a very long time to process for certain large parameters.

This synthetic data request API (which I'll just call "synthetic API" hereafter for brevity) mitigates those problems by exposing an internal HTTP API that can be reached like any other internal HTTP API and by breaking up the data synthesis operations into several asynchronous tasks, some of which run serially but some run highly parallelized.

Data is synthesized in a way that _tries_ to be unique and very _unlikely_ to collide with real-world data. It always creates _new_ users, cloud accounts, images, etc. and _never_ reuses existing records from either real data or previously synthesized data requests. Because the data is being stored and treated like normal customer data, synthesis tries to sidestep collisions by generating random values for IDs and names following conventions that we expect are _not_ used in real-world data. The risk of collision is further mitigated by putting an expiration date-time on all synthesized data and automatically deleting it after that expiration time elapses.

`HTTP POST`s to the synthetic API include various optional values (see `HTTP OPTIONS` for full list of options) such as the numbers of accounts, instances, and runs. The backend immediately stores that request and returns a stub response, and then a sequence of async tasks starts to actually synthesize the data. The response includes a dynamic read-only `is_ready` property; you can `HTTP GET` that response periodically while asnyc tasks run, and it should change from `false` to `true` when all tasks have completed.

Example `HTTP POST` request:

```sh
http localhost:8000/internal/api/cloudigrade/v1/syntheticdatarequests/ \
    cloud_type=aws account_count=1 since_days_ago=30 \
    image_count=10 image_rhel_chance=1.0 \
    instance_count=100 run_count_per_instance_mean=10 hours_between_runs_mean=1
```

Which resulted in this reponse:

```
HTTP/1.1 201 Created
Allow: GET, POST, HEAD, OPTIONS
Content-Length: 473
Content-Type: application/json
Date: Tue, 22 Mar 2022 15:45:32 GMT
Referrer-Policy: same-origin
Server: WSGIServer/0.2 CPython/3.9.6
X-CLOUDIGRADE-REQUEST-ID: f25b90c0-e0ab-49e3-bb2d-c429b8a5f1b1
X-Content-Type-Options: nosniff
X-Frame-Options: DENY

{
    "account_count": 1,
    "cloud_type": "aws",
    "created_at": "2022-03-22T15:45:32.122123Z",
    "expires_at": "2022-03-23T15:45:32.121726Z",
    "hours_between_runs_mean": 1.0,
    "hours_per_run_mean": 2.0,
    "hours_per_run_min": 1.0,
    "id": 1,
    "image_count": 10,
    "image_ocp_chance": 0.5,
    "image_other_owner_chance": 0.5,
    "image_rhel_chance": 1.0,
    "instance_count": 100,
    "is_ready": false,
    "machine_images": [],
    "run_count_per_instance_mean": 10.0,
    "since_days_ago": 30,
    "updated_at": "2022-03-22T15:45:32.122152Z",
    "user": null
}
```

This branch has many of changes, but its commits are all (hopefully) meaningfully incremental and atomic. 🤞


katherine-black (approved):
🎉 👍

Issue #1123, opened 2022-02-16
Add support for ostree-managed filesystems


infinitewarp:
## Summary

As a user running a base image that uses `ostree` to manage the filesystem, I want cloudigrade to identify and report the resulting filesystem as RHEL (if applicable) just like if it was a traditional filesystem layout.

Currently, houndigrade only looks for files relative to the root mount point. With ostree-managed filesystems, we would need to start from wherever the base ostree directory is. We'd also need to add some logic to identify if ostree is being used and what directory should be used as the root.

@infinitewarp recently discovered that some real customers in prod are using images in AWS that appear to be using ostree, and houndigrade+cloudigrade always indicates not-RHEL even though RHEL may have been present deeper in the filesystem.

## Acceptance Criteria

- [ ] Verify that an image using ostree with RHEL in the managed filesystem is correctly identified as having RHEL.
- [ ] Verify that an image using ostree with _no_ RHEL in the managed filesystem is correctly identified as _not_ having RHEL.
- [x] Verify traditional filesystem inspections have no regressions.
- [x] Verify we have added to https://github.com/cloudigrade/houndigrade-test-data new images that would identify as having and not having RHEL within an ostree path.
  - [x] …and it's reasonably small!

## Assumptions and Questions

- Consider testing with one of the community `rhcos` images in AWS. For example, `ami-00058152202db3367`. When mounting a volume of that image into another instance for inspection...
```
# sudo -s
# mkdir /foo4
# mount -t auto -o ro /dev/sdf4 /foo4
# grep "^NAME" /foo4/ostree/deploy/rhcos/deploy/65201d59f5ba7a17ff89b384fdfc959f60b7d41456bca7cacd8dd63f26ba91e2.0/etc/os-release
NAME="Red Hat Enterprise Linux CoreOS"
```
- Do we want to consider CoreOS to "be RHEL" in terms of tracking?
  - If it meets the usual "is this RHEL" criteria, sure!
- Do we want to add a new model attribute to indicate that this is CoreOS and not "normal" RHEL? Or is our current model sufficient? Do we want a new model attribute to indicate that this was discovered under `ostree`?
  - Probably not. That feels too farther into infrastructure that isn't necessarily relevant to "RHEL presence".
- See also https://github.com/ostreedev/ostree and https://ostreedev.github.io/ostree/introduction/

infinitewarp:
@paraggit would you close this issue when you're satisfied that it's covered by tests?

[...]

PR #1154, opened 2022-04-05
Use houndigrade:1.0.0

katherine-black:
#1123 

This will enable cloudigrade to inspect ostree based deployments.


infinitewarp (dismissed):
🚀

Issue #1159, opened 2022-04-08
Capture houndigrade instance logs


katherine-black:
## Summary

As a cloudigrade sysop, I want to have houndigrade stdout/stderr available to me. 

## Acceptance Criteria

- [ ] Verify that we get stdout/stderr output either in cloudwatch or some s3 bucket
  - [ ] And it has a reasonable lifecycle policy. 

## Assumptions and Questions

- Let's consider 90 days for retention/lifecycle/etc.

katherine-black:
Also maybe put the ami name in the inspection json.

katherine-black:
Hey Parag, for this story you should check for a CloudWatch log group created for the ephemeral environment that'll have two log streams in it, one ending in -system and one ending in -inspection.

paraggit:
I have verified that  log stream  ending with '-system' and '-inspection' has created for each ami that has inspected 
 by houndigrade. Following command output is showing that the  log stream has created  successfully, On this observation i am marking this issue as verified and closing this card.
```console 
%> aws logs describe-log-streams --log-group-name houndigrade-ephemeral-ephemeral-79eu2c --query 'logStreams[].logStreamName' --profile smoke
[
    "ami-001bd1d0ff96bf847-inspection",
    "ami-001bd1d0ff96bf847-system",
    "ami-063356dbaf529e4cf-inspection",
    "ami-063356dbaf529e4cf-system"
]

PR #1161, opened 2022-04-18
1159 houndilogs

katherine-black:
#1159 

Demo: https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/houndigrade-ephemeral-ephemeral-thtgmh (in the same account as before)

Logs are split into two files, for readability. All houndigrade inspection logs are in the -inspection stream, the content of /var/log/messages is in -system. Retention period is 14 days.

### DIFF
--- a/cloudigrade/api/clouds/aws/tasks/inspection.py
+++ b/cloudigrade/api/clouds/aws/tasks/inspection.py
@@ -1,4 +1,5 @@
 """Celery tasks related to the AWS image inspection process."""
+import json
 import logging
 
 import boto3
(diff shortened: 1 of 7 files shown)


cloudigrade/api/clouds/aws/tasks/inspection.py
```diff
@@ -111,6 +112,30 @@ def _build_cloud_init_script(ami_id):
...
+        "logs": {
+            "logs_collected": {
+                "files": {
+                    "collect_list": [
+                        {
+                            "file_path": "/var/log/houndigrade",
+                            "log_group_name": f"houndigrade-"
+                            f"{settings.CLOUDIGRADE_ENVIRONMENT}",
+                            "log_stream_name": f"{ami_id}-inspection",
+                            "retention_in_days": 14,
```
infinitewarp, line 125: Did you mean to leave this `14` hard-coded instead of configurable? I actually think 14 days is fine and doubt we'll change it, but whenever I see a magic number, I have to ask at least. 🙂

infinitewarp (dismissed):
Aside from my one inline question, this looks good to me! Nice shell scripting.

cloudigrade/api/clouds/aws/tasks/inspection.py
katherine-black: Hmm, I did miss "Let's consider 90 days for retention/lifecycle/etc." from the original issue. I'll update it to be pulled from a settings env var with a default to 90.

katherine-black: Change is in, let me know what you think. Thank you for the spot!

infinitewarp (approved):

Issue #1135, opened 2022-03-15
Migrate from EBS account_number to org_id


infinitewarp:
This issue is *blocked* until we have a general timeline from sources-api. When will they actually feed us new users/etc. with the org-id? We could start work before sources-api is ready, but if we have already migrated known users and new users aren't including the org-id, we'd have to keep looking up and migrating users as they arrive. Let's try to avoid that complexity by simply waiting until sources-api is done. See sources work here: https://issues.redhat.com/browse/RHCLOUD-18531

## Summary

As a member of the Red Hat console-dot platform, I need to change my primary user identifier from the EBS account_number to the org_id.

## Acceptance Criteria

- [x] Verify cloudigrade appropriately handles auth for incoming HTTP requests that include:
  - [x] ONLY account_number, NOT org_id (current code, likely swatch requests for some time as well)
  - [x] BOTH the org_id, AND account_number (transition?)
    - When we get both, which do we actually use for auth? Do we check both? If we get both, trust the org_id.
  - [x] ONLY the org_id, NOT account_number (future?)
- [x] Verify cloudigrade includes org_id on outgoing requests to sources using:
  - [x] HTTP
  - [x] Kafka

## Assumptions and Questions

- [Slides presented on 2022-03-15 tech talk from Jozef Hartinger](https://docs.google.com/presentation/d/1LsrkKvYQ9sJr0FlNbHLV6OvLYGZbkCgLaL3stLHm-1w/edit#slide=id.g1177c88155f_0_6)
- Today, the incoming 3scale header's account_number *is* the EBS account number (for all intents and purposes).
- We expect that the platform will provide an http service for looking up and translating values for migration.
- Do we need to preserve the old account_number values for any period of time?
  - Yes, we should know both account_number and org_id for some unspecified transition period.
  - This means we should have an initial migration to fetch the org_ids for any existing users.
  - Then we need to retain both the account_number and org_id for new incoming users. The latest 3scale headers should include *both* values for the foreseeable future. Example:
    ```
    {
        "identity": {
            "type": "User"
            "account_number": "901578",
            "org_id": "5318290",
            "user": {
                "first_name": "Jozef",
                "is_active": true,
                "is_internal": true,
                "last_name": "Hartinger",
                "locale": "en_US",
                "is_org_admin": false,
                "username": "jharting",
                "email": "jharting@redhat.com"
            },
            "internal": {
                "auth_time": 0,
                "auth_type": "jwt-auth",
[...]

PR #1192, opened 2022-05-10
Adding support for org_id

abellotti:
Adding support for org-id
   - in Identity header
   - inbound as x-rh-cloudigrade-org-id with PSK's
   - include when sending requests to sources
   - Making the Django auth User's last_name unique since we're using it to stash the user's org_id.
   - Added a django migration from account_number to org_id by invoking the tenant translator.

Testing progress:

- [x] Identity header
- [x] inbound as x-rh-cloudigrade-org-id with PSK's
- [x] include when sending requests to sources
- [x] inbound org-id in kafka message headers from sources
  - [x] org-id via x-rh-identity in header
  - [x] org-id via x-rh-sources-org-id in header
- [x] Django User model migration from account_number to org_id

For https://github.com/cloudigrade/cloudigrade/issues/1135


cloudigrade/api/authentication.py
```diff
@@ -37,19 +37,19 @@ def parse_psk_header(request):
     Get relevant information from the given request's PSK and account number headers.
 
     Returns:
-        (str, str): the tuple of psk and account_number fields.
-            If the psk header is not present and the account_number
-            header is specified, this function returns (None, None).
-            Otherwise it will return (str, str|None) with the valid
-            PSK and the account_number if specified.
+        (str, str, str): the triplet of psk, ord_id and/or account_number fields.
```
infinitewarp, line 40: Even if there are more than two items, it's still called a `tuple`. That's actually the class name.

```
>>> a = (1, 2, 3)
>>> type(a)
<class 'tuple'>
```

abellotti:
Testing progress:

- [ ] Identity header
- [ ] inbound as x-rh-cloudigrade-org-id with PSK's
- [ ] include when sending requests to sources
- [ ] Django User model migration from account_number to org_id

cloudigrade/api/migrations/0059_migrate_account_number_to_org_id.py
```diff
@@ -0,0 +1,56 @@
+# Generated by Django 3.2.12 on 2022-05-06 10:59
+
+from django.db import migrations
+from django.contrib.auth.models import User
+import environ
+import json
+import requests
+
+
+def migrate_account_numbers_to_org_ids(_apps, _schema_editor):
```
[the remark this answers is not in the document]
abellotti: I've added support for post :8000/internal/migrate_account_numbers_to_org_ids (internal api and related celery task and queue) next I will see if I can share/reuse that code with the migration.

abellotti:
With the current PR, anemic tenants are not supported, i.e. org_id and NO account_number, so for example and inbound request let's say from kafka with x-rh-sources-org-id and no x-rh-sources-account-number, will cause the following error:
```
psycopg2.errors.NotNullViolation: null value in column "username" violates not-null constraint
```

as username (account_number) is required, non-unique/indexed and cannot be null.  For supporting anemic tenants, we'd need to migrate the User model to support the non-null pair  (username/last_name) as key, we can also subclass django.contrib.auto.models.User and use that for authentication instead.

cloudigrade/api/tests/tasks/maintenance/test_migrate_account_numbers_to_org_ids.py
```diff
@@ -0,0 +1,153 @@
...
+    tenant_translator_port = "8892"
+    tenant_translator_url = (
+        f"{tenant_translator_scheme}://"
+        f"{tenant_translator_host}:{tenant_translator_port}/"
+        f"internal/orgIds"
+    )
+
+    @mock.patch.dict(os.environ, {"TENANT_TRANSLATOR_SCHEME": tenant_translator_scheme})
+    @mock.patch.dict(os.environ, {"TENANT_TRANSLATOR_HOST": tenant_translator_host})
+    @mock.patch.dict(os.environ, {"TENANT_TRANSLATOR_PORT": tenant_translator_port})
```
infinitewarp, line 32: If we change these to load from `settings` (see other comments), we'll need to updated these decorators to look like:
```
@override_settings(TENANT_TRANSLATOR_SCHEME=tenant_translator_scheme)
```

…and same with the other test functions here.

cloudigrade/api/tests/test_helper.py
```diff
@@ -36,9 +36,9 @@ class SandboxedRestClientTest(TestCase):
 
     def setUp(self):
         """Set up user for the client tests."""
-        self.username = _faker.random_int(min=100000, max=999999)
+        self.account_number = _faker.random_int(min=100000, max=999999)
```
infinitewarp, line 39: I think we should go ahead and add a random `org_id` and use it in the `generate_test_user` call below so we don't forget later when it becomes the primary identifier.

cloudigrade/api/tests/test_models.py
```diff
@@ -193,3 +194,33 @@ def test_synthetic_data_request_post_delete_callback(self):
...
+
+
+class UserModelTest(TestCase):
+    """UserModel tests not covered indirectly via other tests."""
+
+    def test_user_object_string_representation(self):
+        """Test that the string includes the uuid, account_number and org_id."""
+        user_uuid = str(uuid.uuid4())
+        user_account_number = "11111111"
+        user_org_id = "22222222"
```
infinitewarp, line 206: Let's `faker` these values.

infinitewarp (changes requested):
I had to go dig up the Django docs for a refresher on custom User models. 😅 It looks like your changes are pretty spot-on. Good work!

I look forward to another massive cleanup 6-12 months from now when all the dust has settled, everything is consistently using org_id, and we can rip out all the extra account_number handling. That'll be quite the exercise too.

Just a few questions and suggestions inline here…

cloudigrade/api/tests/test_models.py
abellotti: I had the faker calls earlier, but then assertion would fail because it saw Faker objects instead of the strings.  I'll try again.

cloudigrade/api/models.py
```diff
@@ -19,6 +21,83 @@
...
+        max_length=150,
+        unique=True,
+        blank=False,
+        default=user_uuid,
+    )
+    account_number = models.TextField(
+        verbose_name="Account Number",
+        max_length=150,
+        unique=True,
+        null=True,
```
abellotti, line 73: @brad I just remembered I should probably add db_index=True to both account_number and org_id.

cloudigrade/api/tasks/maintenance.py
```diff
@@ -572,3 +577,55 @@ def delete_cloud_accounts_not_in_sources():
...
+
+    """
+    try:
+        env = environ.Env()
+
+        tenant_translator_url = "{}://{}:{}/internal/orgIds".format(
+            env("TENANT_TRANSLATOR_SCHEME"),
+            env("TENANT_TRANSLATOR_HOST"),
+            env("TENANT_TRANSLATOR_PORT"),
+        )
```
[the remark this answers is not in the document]
abellotti: 👍

cloudigrade/api/tests/test_models.py
infinitewarp: Weird… most faker functions should always return common primitives like `int` and `str`. If faker returns an `int`, it would be fine to wrap that in `str(...)` to make any of our type-checking happy.

cloudigrade/api/models.py
infinitewarp: Agreed, good idea!

cloudigrade/api/tests/test_helper.py
abellotti: 👍

cloudigrade/api/tests/test_models.py
abellotti: str() to the rescue 👍

cloudigrade/util/redhatcloud/sources.py
```diff
@@ -329,22 +362,28 @@ def _check_response(error, message):
...
         account_number (str): Account number identifier
+        org_id (str): Org Id identifier
         include_psk (boolean): Whether or not to include the PSK
     """
-    headers = {"x-rh-sources-account-number": account_number}
+    headers = {}
+    if account_number:
+        headers["x-rh-sources-account-number"] = account_number
+    if org_id:
+        headers["x-rh-sources-org-id"] = org_id
```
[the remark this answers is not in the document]
abellotti: Auth to get to this point would have succeeded so either account_number or org_id would be defined. No change w.r.t. failure path here, the error that you mentioned is being trapped by the make_sources_call method.

infinitewarp (approved):
Thanks for addressing all of my questions! I've reviewed the recent changes, and I think it looks good to me.

Before we merge, has @paraggit had a chance to review and play around in an ephemeral environment? Anything that should have new IQE tests before this goes out the door?

abellotti:
Not that I know of, with account_number working as it does now, there shouldn't be an issue. But yea, IQE should be enhanced to include org_id in test requests for the new stuff.  I'm just not when when would this happen timing wise (can tests be added before  it goes out or will that fail IQE) @paraggit any recommendation ?

PR #1211, opened 2022-06-02
change how we handle unreadable S3 files

infinitewarp:
We've seen a rise in permission failures when we try to process CloudTail messages in S3. Until we find and resolve the root cause of those permission problems, we need to stop logging errors (so Sentry calms down) and allow the SQS delivery messages to be deleted.


abellotti (approved):

infinitewarp:
Note: we're merging this without waiting for all checks to be met (codecov/patch specifically) because this is urgent to address a live issue. Discussed in Slack: https://ansible.slack.com/archives/C8VGAPJNN/p1654205422194799?thread_ts=1654197209.421129&cid=C8VGAPJNN

Issue #1209, opened 2022-06-01
Skip S3 files from CloudTrail for unknown accounts


infinitewarp:
## Summary

As a cloudigrade sysop, I do not want to process S3 files from CloudTrail that belong to unknown accounts because they are irrelevant to us. It is currently possible for a customer to set up cloud metering and get a CloudTrail but then break our access to their AWS account in a way that we cannot disable CloudTrail. Until we can better disable CloudTrail in those cases, for now we could at least reduce load on message processing by skipping S3 files that don't have an active cloudigrade CloudAccount.

How do we determine that an S3 file should be skipped? AWS CloudTrail has a very regular pattern in how it writes files into S3. See [Logging with CloudTrail](https://docs.aws.amazon.com/AmazonS3/latest/userguide/cloudtrail-logging.html) and [Identifying Amazon S3 requests using CloudTrail](https://docs.aws.amazon.com/AmazonS3/latest/userguide/cloudtrail-request-identification.html):

> After you set up CloudTrail to deliver events to a bucket, you should start to see objects go to your destination bucket on the Amazon S3 console. These are formatted as follows:
>
> s3://myawsexamplebucket1/AWSLogs/111122223333/CloudTrail/Region/yyyy/mm/dd 

In this example, `111122223333` is the originating customer's AWS account ID. When cloudigrade reads the notification from SQS, we could inspect the S3 file path, parse out the AWS account ID, and confirm that it exists in our database. If it does not exist, we can skip reading that file.

## Acceptance Criteria

- [x] Verify that cloudigrade's `analyze_log` task does not try to read S3 files for inactive accounts.
- [x] Verify that cloudigrade logs an info message when skipping an S3 file.
  - [x] log message should include the whole path so that we may find it later if problems arise
- [x] Verify that normal CloudTrail reading and processing for active accounts is unaffected.

## Assumptions and Questions

- We define "an active cloudigrade CloudAccount" as an account that currently exists, is enabled, not paused, and not synthetic.
- Alternate idea: can we discard SQS notifications about S3 if they originated from an unknown account? Probably not with the options currently available in AWS.
[...]

PR #1233, opened 2022-06-14
Skip aws sqs messages for invalid aws account ids

abellotti:
- Enhancement to skip aws sqs messages that have object references to invalid aws account id's.
- Updated tests so object keys generated are properly formed and referencing actual aws account objects when needed

For issue https://github.com/cloudigrade/cloudigrade/issues/1209


infinitewarp (approved):
Nice and clean! 👍

Issue #1221, opened 2022-06-08
Enable caching with Redis/ElastiCache


infinitewarp:
## Summary

As a cloudigrade developer, I want to be able to cache short-lived values.

The end result is updated Django configs that successfully interface with Redis/ElastiCache for object caching and/or function memoization. Django should support lots of nice cache integrations out of the box. If not, there should be easy to use packages we can install to help.

## Acceptance Criteria

- [x] Verify that caching infrastructure works
- [x] Verify an internal API exists that we can set/check the key/values, and an optional input is the expiration/TTL
- [x] Verify an IQE integration test exists to exercise the new internal API

## Assumptions and Questions

- We believe that we can simply use the Redis/ElastiCache instance that's currently being used by Celery.
- This will be only infrastructure changes plus a proof-of-concept function that gets cached. Actual useful caching will be defined later, such as for https://github.com/cloudigrade/cloudigrade/issues/1214

infinitewarp:
New IQE integration test for the internal cache API exists according to https://ansible.slack.com/archives/C8VGAPJNN/p1656426599473469?thread_ts=1656423852.312429&cid=C8VGAPJNN

That's the last of the ACs. 🎉 Thanks @abellotti and @paraggit !

PR #1240, opened 2022-06-16
Enable Django caching with Redis.

abellotti:
Enable Django caching with Redis.
- include the django-redis package.
- enable the caching against redis using the configured redis host and auth.
- added internal /internal/cache endpoint to getting and setting key values.

For issue: https://github.com/cloudigrade/cloudigrade/issues/1221

### DIFF
--- a/cloudigrade/internal/views.py
+++ b/cloudigrade/internal/views.py
@@ -3,7 +3,8 @@ import logging
 
 from dateutil import tz
 from dateutil.parser import parse
-from django.http import JsonResponse
+from django.core.cache import cache
+from django.http import Http404, JsonResponse
 from django.utils.translation import gettext as _
 from rest_framework import exceptions, permissions, status
 from rest_framework.decorators import (
@@ -305,6 +306,38 @@ def recalculate_concurrent_usage(request):
         )
 
     return Response(status=status.HTTP_202_ACCEPTED)
+
+
+@api_view(["GET", "POST"])
+@authentication_classes([IdentityHeaderAuthenticationInternal])
+@permission_classes([permissions.AllowAny])
+@schema(None)
+def cache_keys(request, key):
+    """Post or Get Cloudigrade cache keys."""
+    method = request.method.lower()
+    if method == "get":
+        """Get a single cache key."""
+        value = cache.get(key)
+        if not value:
+            raise Http404
+        else:
+            return Response(content_type="text/plain", data=value)
+    elif method == "post":
+        """Post a single cache key."""
+        value = request.data.get("value")
+        if not value:
+            raise exceptions.ValidationError(detail="value field is required")
+        timeout = request.data.get("timeout")
+
+        logger.info(
+            _("Setting Cache key='%(key)s', timeout=%(timeout)s"),
+            {"key": key, "timeout": timeout},
+        )
+        if timeout:
+            cache.set(key, value, timeout=int(timeout))
+        else:
+            cache.set(key, value)
+        return Response(data=None)
 
 
 class ProblematicRunList(APIView):
--- a/cloudigrade/internal/views.py
+++ b/cloudigrade/internal/views.py
@@ -24,6 +24,7 @@ from internal.authentication import (
     IdentityHeaderAuthenticationInternal,
 )
 from util import exceptions as util_exceptions
+from util.cache import get_cache_key_timeout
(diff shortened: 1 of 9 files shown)


abellotti:
Adding Django redis caching.

- With a shared Redis, all cache keys are prefixed with "{CLOUDIGRADE_ENVIRONMENT}-cloudigrade"
- Default caching TTL is honored via settings.CACHE_TTL_DEFAULT (60 seconds)
- Internal API exposed for setting and querying cache keys.


```
$ http post localhost:8000/internal/testkey/ value="test value" timeout=120

POST /internal/cache/testkey/
{
    value: "test value",
    timeout: "120",
}
```

```
$ http localhost:8000/internal/testkey/

GET /internal/cache/testkey/
```

cloudigrade/internal/views.py
```diff
@@ -307,6 +308,38 @@ def recalculate_concurrent_usage(request):
...
+def cache_keys(request, key):
+    """Post or Get Cloudigrade cache keys."""
+    method = request.method.lower()
+    if method == "get":
+        """Get a single cache key."""
+        value = cache.get(key)
+        if not value:
+            raise Http404
+        else:
+            return Response(content_type="text/plain", data=value)
```
infinitewarp, line 324: One maybe not so minor request: How would you feel about including the key's TTL in the response as an HTTP response header, or is it even possible? Ideally something like…

```
ttl = which function call?
headers = {"X-CLOUDIGRADE-CACHE-TTL": ttl}
...
return Response(content_type="text/plain", headers=headers, data=value)
```

I looked briefly through the standard Django docs for the cache framework, and I didn't find any functions that return the TTL/expiration time. This is a little surprising, but I guess maybe it doesn't exist if all backends don't support explicit timing.

I did find these suggestion in [this stack overflow question](https://stackoverflow.com/questions/18203116/can-i-get-expiration-time-of-specified-key-in-django-cache), though:
```
# to get the unix timestamp
cache._expire_info.get('foo') 
```
or for Redis specifically:
```
cache.ttl('foo:bar:2020-09-01')
```
Though, the latter probably wouldn't work with our unit tests sine we use locmem.

All of this makes me think that what I'm asking for either isn't possible at all or probably isn't possible using supported/stable APIs. 😕 

Thoughts?

infinitewarp (changes requested):
Looking good. 👍  I suspect the answer to my one question here will be "no…", and if that's the case, I'll just approve what we have now.

cloudigrade/internal/views.py
abellotti: it is definitely possible to get the ttl in either case as per the get_key_expiration() example in that   stack overflow. So if you think we need it, let's return that too. The only thing I'd recommend differently is the response, instead of the data in body and ttl in header (2 places to look at), what about returning a json, i.e. similar to the post data above https://github.com/cloudigrade/cloudigrade/pull/1240#issuecomment-1157640881 for consistency, we can return the equivalent in json. so you'd get the data via response.data["value"].

```
$ http localhost:8000/internal/testkey/

GET /internal/cache/testkey/
{
    value: "test value",
    timeout: "120",
}
```

thoughts ?

abellotti: @infinitewarp ☝️ is not too bad: https://github.com/abellotti/cloudigrade/compare/435727a56a7f45dc2d701389161264f25d16fa24...cb8565d4f64f59f025004e0293780b3bcc8ec1d6

Let me know

infinitewarp: Sorry for the delay! Yes, your sample output looks like a good solution to me. 🙂 

Your new `get_cache_key_timeout` is a little more complex than I anticipated, but it's only as complex as necessary now that I see all the moving parts. I'm okay with this!

I would imagine it's very unlikely that we will change to a different cache backend any time soon, and if we do, we'll need to update that function since new backend-specific failures likely won't be caught by unit tests. Maybe we should work with @paraggit to add an IQE test that exercises the new internal API to verify in the ci.ext PR check that it always behaves as expected. Thoughts? Other than that, I think this is good to go.

abellotti: Thanks. I agree on tests. While the unit tests will exercise the LocMemCache, Parag's ci.ext will exercise the Redis side of things.

infinitewarp (approved):
👍

Issue #1232, opened 2022-06-13
Implement DVO overrides for replica counts


infinitewarp:
## Summary

As a cloudigrade sysop, I want to override SRE's DVO checks that enforce 3+ replicas for each deployment because some of my deployments (`cloudigrade-beat`, `cloudigrade-listener`) require only 1 replica.

Current related firing issues:
- [CLOUDMETER-9](https://issues.redhat.com/browse/CLOUDMETER-9) deployment_validation_operator_minimum_three_replicas crcs02ue1
- [CLOUDMETER-13](https://issues.redhat.com/browse/CLOUDMETER-13) deployment_validation_operator_minimum_three_replicas crcp01ue1

DVO Dashboards:
- [stage](https://grafana.app-sre.devshift.net/d/dashdotdb/dash-db?orgId=1&var-datasource=dashdotdb-rds&var-cluster=crcs02ue1&var-namespace=cloudigrade-stage)
- [prod](https://grafana.app-sre.devshift.net/d/dashdotdb/dash-db?orgId=1&var-datasource=dashdotdb-rds&var-cluster=crcp01ue1&var-namespace=cloudigrade-prod)

What is DVO? [See Deployment Validation Operator docs in app-interface](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/dvo.md).
> The [Deployment Validation Operator](https://github.com/app-sre/deployment-validation-operator) (DVO) checks deployments and other resources against a curated collection of best practices. Under the hood, it continuously runs the static analysis tool [KubeLinter](https://github.com/stackrox/kube-linter).

What is our problem? From [Steve's comment here](https://docs.google.com/spreadsheets/d/1mgiQA2dG4bnijCTlaBmKC5Ns7-AjMiPVhStQQokWtz0/edit?disco=AAAAYemoBR4):
> DVO checks will fail, so you'll end up needing to add annotations as per the instructions for "specific workloads" […] 
> […] generally you don't want to ignore the check at the app level (app-1.yml)
> This isn't a blocker in the current state, but we will end up checking DVO alerts again in the checkpoint before going to production. I'd still suggest continuing to follow-up on support for this in Clowder.

Where is Clowder's support?
- [2022-06-01 Slack thread](https://coreos.slack.com/archives/C022YV4E0NA/p1654107782438899)
[...]

PR #1246, opened 2022-06-23
Add overrides for DVO

abellotti:
- Add override for DVO so that the minimum check of 3 replicas is not enforced for cloudigrade-beat
and cloudigrade-listener where they must run with just 1 replica.

For: https://github.com/cloudigrade/cloudigrade/issues/1232

### DIFF
--- a/deployment/clowdapp.yaml
+++ b/deployment/clowdapp.yaml
@@ -328,12 +328,12 @@ objects:
       webServices:
         public:
           enabled: false
-      metadata:
-        annotations:
-          ignore-check.kube-linter.io/minimum-three-replicas: "cloudigrade-beat must only run with 1 replica"
       podSpec:
         image: ${IMAGE}:${IMAGE_TAG}
         imagePullPolicy: Always
+        metadata:
+          annotations:
+            ignore-check.kube-linter.io/minimum-three-replicas: "cloudigrade-beat must only run with 1 replica"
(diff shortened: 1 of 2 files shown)


abellotti:
We're kind'a blocked and waiting for https://issues.redhat.com/browse/RHCLOUD-20060 so Clowder supports this.

infinitewarp (changes requested):
Adding a note in here so we don't forget…

We may need to put similar annotations on [stage-smoke.yaml](https://github.com/cloudigrade/cloudigrade/blob/master/deployment/stage-smoke.yaml) because DVO in stage is reporting failures for `deployment_validation_operator_no_liveness_probe` and `deployment_validation_operator_no_readiness_probe`. See "Validation Details" on [this dsahboard](https://grafana.app-sre.devshift.net/d/dashdotdb/dash-db?orgId=1&var-datasource=dashdotdb-rds&var-cluster=crcs02ue1&var-namespace=cloudigrade-stage) to confirm.

_Or_ we need to reach out to the QE org to get them make whatever changes are necessary so that we don't need to ignore those checks on the smoke jobs.

infinitewarp (approved):
Looks good to me! 🎉 Here's hoping it works as expected in stage and prod! 🤞🍀

infinitewarp (approved):

PR #1252, opened 2022-06-28
Adding the TENANT_TRANSLATOR_* env's for the worker too

abellotti:
- Adding the TENANT_TRANSLATOR_* env's for the worker too.
- Moving the declaration of those into the shared section.
- Logging the TENANT_TRANSLATOR_* entries to help in debugging.


infinitewarp (approved):
👍

PR #1253, opened 2022-06-30
restore "username" filtering on internal users api

infinitewarp:
We removed the ability to filter the internal "users" API by the "username" query string argument in https://github.com/cloudigrade/cloudigrade/pull/1192, but we recently learned that the swatch service was relying on that. This change reimplements that filter by filtering on User.account_number with the given "username" and adds several unit tests to ensure we don't accidentally lose this in the future.

* allow filtering internal users API by "username" (which actually filters "account_number")
* add "username" to internal users API responses
* allow filtering internal users API by "uuid"

### DIFF
--- a/cloudigrade/internal/filters.py
+++ b/cloudigrade/internal/filters.py
@@ -64,6 +64,21 @@ class InternalInstanceEventFilterSet(django_filters.FilterSet):
         }
 
 
+class InternalUserFilterSet(django_filters.FilterSet):
+    """FilterSet for limiting api.User for the internal API."""
+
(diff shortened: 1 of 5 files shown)


abellotti (approved):

infinitewarp:
IQE (`ci.ext.devshift.net PR build`) appears to be failing for reasons unrelated to this change; it is also failing on other branches that have zero code differences compared to master. Since this is a critical fix that we've tested manually and verified with unit tests and code review, @abellotti and @gtanzillo agreed we should override to merge without that PR check.

:shipit:

Issue #1229, opened 2022-06-13
Support managed Kafka with authentication


infinitewarp:
## Summary

As a cloudigrade sysop, I need to support the new managed Kafka with authentication because the platform says this is required by 2022-08-31.

- [Upcoming migration to Managed Kafka - Tenant Actions Required](https://source.redhat.com/groups/public/consoledot/consoledot_blog/upcoming_migration_to_managed_kafka_tenant_actions_required)
  > At the end of August 2022, Platform services that currently use the platform-mq-* Kafka resources will need to be shifted to new Kafka clusters [...] Applications that cannot currently authenticate with Kafka must make changes to their Kafka connection configuration no later than August 31, 2022.
- [RHCLOUD-19432 Application Updates for RHOSAK/Kafka Auth](https://issues.redhat.com/browse/RHCLOUD-19432)
- [RHCLOUD-19698 Update cloudigrade to use Kafka authentication](https://issues.redhat.com/browse/RHCLOUD-19698)
  > In preparation for the move to RHOSAK/Managed Kafka, all services that use Kafka need to be able to authenticate to the brokers with a user id and token provided either through an OpenShift secret or from Clowder.
- [Previous Document Detailing KafkaAuth Setup](https://docs.google.com/document/d/1X0Sgj9Wg3YxmIcRd9-pAaPiJ2r3nIT06U5HBvp5q0oA/edit#heading=h.cbh7v879kfn2)
  > The Kafka CA cert and User credentials  will be provided in the cdappconfig.json
  > [Example of how ingress pulls config from Clowder](https://github.com/RedHatInsights/insights-ingress-go/blob/bbfee1e4b3a8cc94c383f261d4d039b9bd0756e9/internal/config/config.go#L122)
  > Confluent Python: [Example from Puptoo](https://github.com/RedHatInsights/insights-puptoo/pull/160)
  >  NOTE: If you already made Kafka changes using the previous instructions, you will need to update your code to set a username, password,sasl_mechanism and also update your ssl protocol setting as shown in the example above.

## Acceptance Criteria

- [ ] Verify we understand clearly how the migration/switch will occur
  - Update this issue and/or share with the team in Slack what you find.
- [x] Verify deployments continue to work with Kafka (no regressions!)
- [x] Verify [RHCLOUD-19698 Update cloudigrade to use Kafka authentication](https://issues.redhat.com/browse/RHCLOUD-19698) is resolved

## Assumptions and Questions

[...]

PR #1257, opened 2022-07-01
Adding support for Managed Kafka

abellotti:
Adding support for Managed Kafka
- Updated dependency of app-commonp-python that supports managed kafka.
- Updated the LISTENER_SERVER and LISTENER_PORT configuration variable to KAFKA_SERVER_HOST and KAFKA_SERVER_PORT as they are used for also sending kafka messages and not just by the listener.
- Honoring the Kafka server CA Location and SASL username and password if specified in the clowder configuration.

For: https://github.com/cloudigrade/cloudigrade/issues/1229


infinitewarp (approved):
:+1:

PR #1271, opened 2022-07-22
pr_check.sh: stop using deprecated value format

mirekdlugosz:
During demo done by Parag, I noticed deprecated warning in pr-check job logs:

    --set-template-ref: <app>/<component>=<ref> syntax is deprecated, use <component>=<ref>

This change was introduced back in 2021 in https://github.com/RedHatInsights/bonfire/pull/44/ . Component names are unique across the whole config, so there's no point in namespacing components by app names - `foo/foo=bar` and `foo=bar` are supposedly equivalent.

I guess CI results from this PR will tell if this change is really safe to merge.

### DIFF
--- a/deployment/scripts/pr_check.sh
+++ b/deployment/scripts/pr_check.sh
@@ -43,8 +43,8 @@ bonfire deploy \
     ${APP_NAME} \
     --source=appsre \
     --ref-env insights-stage \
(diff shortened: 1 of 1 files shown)


mirekdlugosz:
Yup, this particular warning is no longer in logs.
Otherwise, seems ok. Took about the same time as usual, no failures reported.

infinitewarp (approved):

PR #1279, opened 2022-07-27
Defining the Local MemCache as the Default Django Cache.

abellotti:
- Using the Memory Cache as the default Django cache.


abellotti:
With this PR, both cache can be accessed separately as follows:

```
from django.core.cache import cache

mem_cache = caches["default"]
redis_cache = caches["redis"]

mem_cache.set("test", "In Memory Key")
redis_cache.set("test", "Redis Key")

mem_cache.get("test")  
"In Memory Key"

redis_cache.get("test")
"Redis Key"
```

PR #1281, opened 2022-07-28
add db index for created_at and updated_at on all models

infinitewarp:
I noticed that we don't have indexes on *any* `created_at` or `updated_at` columns, but there are several places we filter and order by those values on large tables that would probably benefit from indexes.


abellotti:
Nice find and update, should help with our performance. 👍

PR #1282, opened 2022-07-28
make cache_memoize calls use locmem

infinitewarp:
This really shouldn't be necessary, but we're having redis issues and performance issues, and I'm wondering if they're related, and maybe using redis less will help.

feelsbadman.jpg


infinitewarp:
Tested locally by running `redis-cli monitor` while executing specific commands previously known to talk to redis including these cached functions. Calls to these functions did nothing to the monitor. Calls to other functions like `check_and_cache_sqs_queues_lengths` and the custom `metrics` functions continued using redis normally.

PR #1283, opened 2022-07-28
set custom visibility timeout on cloudtrail event queue

infinitewarp:
This appears to be necessary because our cloudtrail event processing sometimes takes longer than 30 seconds, and AWS's default visibility timeout of 30 seconds means that some messages will unexpectedly be requeued (and then reprocessed later as well).


gtanzillo (approved):
LGTM 👍

Issue #1287, opened 2022-08-01
remove org admin requirement


infinitewarp:
## Summary

As a consoledot user, I want to use cloudigrade without having the org admin flag set because that option has been deprecated and I may not be able to set it.

At the very least, we need to relax access to sysconfig so non-admin users can follow the UI flow for creating sources.

## Acceptance Criteria

- [ ] Verify that a Red Hat user without org admin successfully retrieves a normal sysconfig response.

## Assumptions and Questions

- See also https://issues.redhat.com/browse/CLOUDMETER-16
- Eventually we should integrate with RBAC and restrict access to specific roles, but that's not important now.
- Relaxing controls to make APIs more "public" is okay because we still require the user to be logged in via Red Hat SSO.

paraggit:
We Have changed our tests users and removed org_admin role from it. Tests are passed without failures. No  issue seen in tally sync job  since the changes has merged in prod environment . In these basis this issue has verified and we are moving to closed state.

PR #1293, opened 2022-08-03
identity header is_org_admin is no longer required to access any APIs

infinitewarp:
Also clean up some docstrings and rename require_account_number to require_account_number_or_org_id to be more accurate.

For https://github.com/cloudigrade/cloudigrade/issues/1287


infinitewarp:
I don't understand why the GitHub actions for our `py39` jobs are taking so long on this PR. Running locally with redis and postgresql _stopped_ and disconnected from the internet (just to make sure it's not trying to call out to other services), `tox -e py39` just now completed successfully in only 54 seconds. 🤔

abellotti (approved):

Issue #1267, opened 2022-07-20
Internal API to keep persistent users


infinitewarp:
## Summary

As a cloudigrade developer and sysop, I want to have *some* users remain persistent and be ignored by the usual cleanup processes so that I can more easily make internal and in-development API requests.

Let's add a "permanent" flag to the user model and allow changing its value via an internal API. We already have an internal API that lists users; we could extend its functionality to allow put/post/patch to update the user.

Let's also allow creation and deletion via the same internal user API, but only allow a delete request to complete if the user has no cloud accounts.

## Acceptance Criteria

- [x] Verify the internal user API allows us to create a user
- [x] Verify the internal user API allows us to delete a user if the user has no cloud accounts
- [x] Verify the internal user API does not allow us to delete a user if the user has cloud accounts (400 Bad Request with appropriate message)
- [x] Verify the internal user API allows us to set/unset user permanence
- [x] Verify that when permanent is true, the user object still exists and is usable for API requests after its cloud accounts are deleted
- [x] Verify that when permanent is false, the user object is deleted like it normally has been

## Assumptions and Questions

- This could be a boolean flag on our new user model.
- This would need to be excluded from at least two places: 1) when we auto-delete users upon deleting accounts, and 2) the scheduled job that deletes any "orphaned" users.

abellotti:
With #1295 the following are actual examples via HTTPie:

- Creating a new user with an account_number and org_id

```
$ localhost:8000/internal/api/cloudigrade/v1/users/ account_number=10001 org_id=20001
```

```
HTTP/1.1 201 Created
Allow: GET, POST, HEAD, OPTIONS
Content-Length: 181
Content-Type: application/json
Date: Thu, 04 Aug 2022 14:15:31 GMT
Referrer-Policy: same-origin
Server: WSGIServer/0.2 CPython/3.9.7
X-CLOUDIGRADE-REQUEST-ID: e02afb4e-21d1-4d48-9416-e04b4b44bba8
X-Content-Type-Options: nosniff
X-Frame-Options: DENY

{
    "account_number": "10001",
    "date_joined": "2022-08-04T14:15:31.376080Z",
    "id": 15,
    "is_permanent": false,
    "org_id": "20001",
    "username": "10001",
    "uuid": "bd3554cf-8af8-4c97-8334-272094bd1bb2"
}
```

- Updating that user to have the permanent flag set:

```
[...]

PR #1295, opened 2022-08-04
Adding support for permanent users

abellotti:
- Updated Api migrations to add the is_permanent boolean, defaults to False. Tested both forward and reverse migrations.
- Updated internal serializer to include is_permanent.
- Updated related api doc examples
- Migration scripts dealing with the api.User model should be accessing the historical model and not the current model. Updated 0062_migrate_account_number_to_org_id to use the get_model method. this was causing a stack trace during migration with the new is_permanent flag added.
- Implemented the Internal User DELETE function to make sure we  don't delete a user if it belongs to cloud accounts. If it does not, then trigger a force delete.
- Adding ability to delete users, honor the is_permanent flag in the model's new delete() function and support a force=True option.
- Support creating new users (POST), updating current users (PUT/PATCH), and deleting users (DELETE)

For: https://github.com/cloudigrade/cloudigrade/issues/1267


cloudigrade/api/models.py
```diff
@@ -92,12 +95,19 @@ def __str__(self):
         """Return the friendly string representation for the Api User."""
         return (
             f"{self.__class__.__name__}("
+            f"id={self.id}, "
```
infinitewarp, line 98: Want to add `is_permanent` here too, or was it an intentional omission?

cloudigrade/internal/viewsets.py
```diff
@@ -93,13 +93,33 @@ def perform_destroy(self, instance):
...
 
 
-class InternalUserViewSet(InternalViewSetMixin, viewsets.ReadOnlyModelViewSet):
-    """Retrieve or list Users for internal use."""
+class InternalUserViewSet(
+    InternalViewSetMixin,
+    mixins.CreateModelMixin,
+    mixins.UpdateModelMixin,
+    mixins.DestroyModelMixin,
+    viewsets.ReadOnlyModelViewSet,
```
infinitewarp, line 101: I think this could be simplified since it's no longer read-only:
```suggestion
    viewsets.ModelViewSet,
```
See also https://github.com/encode/django-rest-framework/blob/master/rest_framework/viewsets.py#L235

cloudigrade/api/models.py
abellotti: 👍 no problem I'll add it. we'll also add date_joined for consistency with the InternalUserSerializer.

cloudigrade/internal/viewsets.py
abellotti: 👍

infinitewarp (approved):

PR #1296, opened 2022-08-04
Fixed the examples in docs/rest-api-examples.rst

abellotti:
- Fixed the examples for docs/rest-api-examples.rst where we had the older /api/v2/\<entity\> urls that did not include the service name. should be /api/cloudigrade/v2/\<entity\>.


infinitewarp (approved):
Good catch. Thanks!

PR #1300, opened 2022-08-09
raise error when creating AzureCloudAccount if its subscription ID is already used

infinitewarp:
for https://github.com/cloudigrade/cloudigrade/issues/1262


abellotti (approved):
LGTM!! Thanks for also adding the test.

PR #1301, opened 2022-08-09
replace old "clount" nickname with "cloud account"

infinitewarp:
Long ago, we sometimes referred to cloud accounts as "clounts" in some conversational and shorthand contexts, but now its occasional use is inconsistent with the rest of our code, and its meaning is not obvious. I ran into this again while updating some old tests, and I decided it's finally time to purge this old slang.


gtanzillo (approved):
Wow, it was in a lot of places! LGTM 👍

abellotti (approved):

PR #1303, opened 2022-08-10
apparently the kafka cert is not available everywhere and we need to allow sasl and users *without* the cert being present

infinitewarp:
Why?

https://ansible.slack.com/archives/C022YV4E0NA/p1660144914692639?thread_ts=1660140332.243559&cid=C022YV4E0NA
> managed kafka doesn't have a self-signed cert to use so we just don't set it

### DIFF
--- a/cloudigrade/config/settings/base.py
+++ b/cloudigrade/config/settings/base.py
@@ -666,12 +666,12 @@ if isClowderEnabled():
         KAFKA_SERVER_CA_LOCATION = clowder_cfg.kafka_ca()
         __print("Clowder: Kafka server CA defined")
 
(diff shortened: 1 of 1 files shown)


gtanzillo (approved):
LGTM 👍

abellotti:
Looks like things will work as per the method: https://github.com/cloudigrade/cloudigrade/blob/ebdcc1a061038fa411e5777e7568002e4ab0f925/cloudigrade/util/redhatcloud/sources.py#L424,  though would have been nice to test first on Ephemeral, but that's not the case now.

abellotti (approved):

infinitewarp:
related to https://github.com/cloudigrade/cloudigrade/issues/1229

PR #1304, opened 2022-08-10
load kafka topic names from clowder config

infinitewarp:
why?

because apparently sometimes the topic names aren't what we think they are.

https://ansible.slack.com/archives/C022YV4E0NA/p1660147166446529?thread_ts=1660140332.243559&cid=C022YV4E0NA

> In MK the topic names have an appended piece, the topic there would be platform-mq-stage.platform.sources.event-stream Clowder should be providing the updated topic names so long as you're pulling the config from there


abellotti (approved):
LGTM!!

infinitewarp:
related to https://github.com/cloudigrade/cloudigrade/issues/1229

PR #1309, opened 2022-08-11
download and md5sum the postgresql repo rpm for diagnostic purposes

infinitewarp:


Dockerfile
```diff
@@ -3,7 +3,11 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6 as base
 
 WORKDIR /opt/cloudigrade
 
-RUN rpm -iv https://download.postgresql.org/pub/repos/yum/reporpms/EL-8-x86_64/pgdg-redhat-repo-latest.noarch.rpm \
+RUN curl -so 'pgdg-redhat-repo-latest.noarch.rpm' 'https://download.postgresql.org/pub/repos/yum/reporpms/EL-8-x86_64/pgdg-redhat-repo-latest.noarch.rpm' \
+    && md5sum 'pgdg-redhat-repo-latest.noarch.rpm' \
+    && rpm --verbose -K 'pgdg-redhat-repo-latest.noarch.rpm' || true
+
+RUN rpm -iv 'pgdg-redhat-repo-latest.noarch.rpm' \
```
abellotti, line 10: do we need to delete that rpm file once done with it ?

infinitewarp: We could, but it's only a few KB, and I'd like to keep it around so we can look at it _directly_ if we run into the weird signing issues again.

abellotti: 👍

abellotti (approved):

PR #1313, opened 2022-08-15
update dependencies

infinitewarp:
[Bump nbconvert from 5.5.0 to 6.3.0](https://github.com/cloudigrade/cloudigrade/pull/1306) and [Bump mistune from 0.8.4 to 2.0.3](https://github.com/cloudigrade/cloudigrade/pull/1310) cannot coexist due to conflicting version requirements.

This PR updates various _other_ dependencies and rebuilds the lock to find the latest resolvable versions for `mistune` and `nbconvert`. The result is `nbconvert` wins, and we're stuck with this problem: [Mistune v2.0.2 vulnerable to catastrophic backtracking](https://github.com/cloudigrade/cloudigrade/security/dependabot/6).

Fortunately, both `nbconvert` and `mistune` appear to be required by only `jupyter` which is a dev dependency that we _do not_ normally build into our images. So, the potential risk should be limited to local development PEBCAK 🙂 and I'm willing to ignore that security alert (for now).


gtanzillo (approved):
LGTM 👍

Issue #1243, opened 2022-06-16
Azure: Disconnect Lighthouse from customer subscription


infinitewarp:
## Summary

As a cloudigrade user via sources, when I destroy my Azure source, I want cloudigrade to disconnect and terminate its relationship with my Azure subscription.

We currently establish trust between ours and the customer's Azure subscriptions via Lighthouse, but we never break that relationship. We need to figure out how to do that, and do it whenever the source is gone! (Just like destroying the AWS CloudTrail.)

## Acceptance Criteria

- [x] Ideally verify that deleting a source breaks our access to the customer's subscription
  - If this is not technically possible from our side, we must have explanations documented.
- [x] Verify answers are documented for the questions below
- [x] Verify we at least have instructions so that we as cloudigrade devs can disconnect our subs from each other so we can cleanly start over with sources.

## Assumptions and Questions

- Is this possible to perform from within our cloudigrade subscription?
- Do we have to reach into the customer's subscription somehow? Do we have enough access within our relationship to do that?
- Do we have to ask the customer to do something?
- Is the relationship effectively established forever and we just have to ignore their subscription?

abellotti:
First, establishing the relationship between our cloudigrade azure account and the customer's tenant. In doing this development, I want to be able to work in the Ephemeral environments before anything is merged/pushed to stage/prod. 

Based on https://github.com/cloudigrade/cloudigrade/issues/1026#issuecomment-1010386770 Here are updated instructions to work in Ephemeral.

_Note_: Since we do not have access to Ephemeral externally, we need to manually upload the Microsoft.Template for lighthouse.

First define the AZURE environment variables to your enterprise app, i.e. <login>-personal-dev:

```
export AZURE_TENANT_ID = <Tenant/Directory Id of Cloudigrade - Ephemeral>
export AZURE_SUBSCRIPTION_ID = <Subscription ID of Cloudigrade Ephemeral>
export AZURE_CLIENT_ID = <Enterprise Application (Client) ID for <login>-personal-dev>
export AZURE_SP_OBJECT_ID = <Object ID of Enterprise Application for <login>-personal-dev>
export AZURE_CLIENT_SECRET = <Client Secret for the App Registration for <login>-personal-dev>
```

Next, deploy Cloudigrade to your Ephemeral namespace via the `poetry run ansible-playbook ... deployment/playbooks/manage-clowder.yml` command.

In the following examples, we first port forward 8000 to Cloudigrade and 8002 to Sources:

[...]

PR #1319, opened 2022-08-22
Enable Azure lighthouse  registration assignment deletes

abellotti:
- Adding the Azure Managed Services registration assignment delete role id to the lighthouse azure offer template, so service principals as referenced by the AZURE_SP_OBJECT_ID can remove a lighthouse customer delegation via the service provider's Azure portal UI.

For #1243

### DIFF
--- a/cloudigrade/util/azure/__init__.py
+++ b/cloudigrade/util/azure/__init__.py
@@ -8,15 +8,28 @@ from util.azure.identity import (
 
 
 # Azure Permissions
+#
 # Azure permissions or roles are defined via their object ids, this
 # unfortunately has the side effect of making it very hard to reference
 # what a specific id is, this section is a way to map these azure ids
 # to a developer readable name to be used in our template.
-# Reader - acdd72a7-3385-48ef-bd42-f606fba81ae7
+#
+# The role definition id's referenced here below are Azure
+# built-in roles. The id's are static in nature and are documented
+# here: https://docs.microsoft.com/en-us/azure/role-based-access-control/built-in-roles
+#
+# The Reader role allows us to view all resources as needed when doing the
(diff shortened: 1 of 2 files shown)


cloudigrade/util/azure/__init__.py
```diff
@@ -15,6 +15,10 @@
 # Reader - acdd72a7-3385-48ef-bd42-f606fba81ae7
 AZURE_READER_ROLE_ID = "acdd72a7-3385-48ef-bd42-f606fba81ae7"
 
+# Managed Services Registration assignment
+# Delete Role - 91c1777a-f3dc-4fae-b103-61d183457e46
+AZURE_MS_REG_ASSIGNMENT_DELETE_ROLE_ID = "91c1777a-f3dc-4fae-b103-61d183457e46"
```
infinitewarp, line 20: Should this (and the other hard-coded IDs above, now that I think about it) be loaded in from environment variables?

Will this be different for ephemeral vs. stage vs. prod? Or is this a truly universal ID across Azure? If the former, how did you create it? If the latter, where did it come from?

abellotti: this are static as they are Azure built-in role id's, they don't change.  Here's the list of them, https://docs.microsoft.com/en-us/azure/role-based-access-control/built-in-roles  (just a few).

infinitewarp: Because I'm sure to forget this, would you add a comment above these variables with that link and a succinct explanation?

infinitewarp: Excellent comment additions. Thank you! 🎉

infinitewarp (approved):

infinitewarp (approved):

PR #1347, opened 2022-09-28
report error if subscription ID is not valid UUID

infinitewarp:
for https://github.com/cloudigrade/cloudigrade/issues/1340

### DIFF
--- a/cloudigrade/api/clouds/azure/util.py
+++ b/cloudigrade/api/clouds/azure/util.py
@@ -1,5 +1,6 @@
 """Utility functions for Azure models and use cases."""
 import logging
+import uuid
 
 from django.db import IntegrityError, transaction
 from django.utils.translation import gettext as _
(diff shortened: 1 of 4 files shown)


cloudigrade/api/clouds/azure/util.py
```diff
@@ -72,9 +73,15 @@ def create_azure_cloud_account(
...
+        try:
+            subscription_id_as_uuid = uuid.UUID(subscription_id)
+        except ValueError:
+            error_code = error_codes.CG1006
+            error_code.notify(user.account_number, user.org_id, platform_application_id)
+            raise ValidationError({"subscription_id": error_code.get_message()})
         try:
             azure_cloud_account = AzureCloudAccount.objects.create(
-                subscription_id=subscription_id
+                subscription_id=subscription_id_as_uuid
```
abellotti, line 84: why the shift to using subscription_id_as_uuid (now a UUID object) here instead of subscription_id (str)?

infinitewarp: `AzureCloudAccount.subscription_id` is already defined as a `models.UUIDField`, and that field happily takes a UUID-like string (old code) or a proper `uuid.UUID` object (new code). If it _doesn't_ get a proper `uuid.UUID`, internally it converts the input to a `uuid.UUID`. That internal conversion is where things were exploding in the error reported by Sentry when given junk inputs.

So, since I'm already casting the string to a proper `uuid.UUID` a few lines above in order to sanity-check its contents, passing that new `uuid.UUID` object into the function here both works and skips an unnecessary/duplicate conversion deeper within the `models.UUIDField` class. It's a micro-optimization, for sure, but an easy one.

abellotti (approved):

PR #1353, opened 2022-10-04
fix incorrect type match for Azure subscription IDs

infinitewarp:
This fixes an issue where `AzureCloudAccount.enable()` might never work after the initial creation of the `AzureCloudAccount` object. `AzureCloudAccount.subscription_id` returns a `uuid.UUID`, but the return value from `get_cloudigrade_available_subscriptions` was a list of plain `str`s. This means that future calls to `enable` would never succeed and would always produce an error.

This fix simply changes that returned list to include `uuid.UUID` objects instead of `str`s. In general, when we compare UUIDs, we should be comparing `uuid.UUID` _objects_ since equivalent UUIDs could have strings formatted differently and fail equivalency checks.

I added a type hint to `get_cloudigrade_available_subscriptions`, but we don't enforce type checking anywhere yet. Maybe some day in the future we will. For now, it's just for our IDEs to remind us in development.

Also, as I was reviewing code around Azure and UUIDs for this change, I also spotted two nearby `logger.warn`s that I corrected to `logger.warning` because `warn` is an old deprecated method.

### DIFF
--- a/cloudigrade/api/clouds/azure/models.py
+++ b/cloudigrade/api/clouds/azure/models.py
@@ -279,7 +279,7 @@ def delete_lighthouse_registration(*args, **kwargs):
             id_list = reg_assignment.id.split("/")
             if state == "Succeeded" and "/".join(id_list[1:3]) == tenant_scope:
                 if registration_name:
-                    logger.warn(
+                    logger.warning(
                         _(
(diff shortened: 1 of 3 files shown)


gtanzillo (approved):
LGTM 👍

PR #1367, opened 2022-10-18
ISSUE-1206 Update Azure Instances Hourly

BrittRH:
# Description

Cloudigrade should know about instances starting/stopping after it has already been defined by sources. This should get the list of instances that cloudigrade already knows about and adds "power on" and "power off" events to our DB.

# Solution
- I added a celery task that defaults to running hourly for updating all of the Azure instances.
- Modified the `create_initial_azure_instance_events` function to call `save_instance_events` even when the VMs are not running so you always get the initial event
- Modified the `save_instance_events` function to create InstanceEvents for both `power_on` and `power_off` based on if the VM is running or not

To Do:

- [x] Linting
- [x] Testing (adding test coverage)

# Testing

## Setup
I've set up my environment locally to test, following the instructions here: https://github.com/cloudigrade/cloudigrade/wiki/How-do-I-run-Sources-&-Kafka-locally-for-development. You will also need to have redis, and postgres running locally, and the various celery related commands.

All in all, you'll need terminal windows open with the following:
- cloudigrade `runserver`
- cloudigrade: `./manage.py listen_to_sources`
- cloudigrade: `celery --app cloudigrade.config worker --loglevel info --pool solo --without-gossip --without-mingle --concurrency 1 --task-events --queues $CELERY_QUEUE_NAMES`
- cloudigrade: `UPDATE_AZURE_INSTANCE_EVENTS_SCHEDULE=30 celery --app cloudigrade.config beat --loglevel info --scheduler django_celery_beat.schedulers:DatabaseScheduler`
- sources (from wiki)
- kafka & zookeeper (from wiki)

**NOTES**
The task by default runs every hour, however, I've added the `UPDATE_AZURE_INSTANCE_EVENTS_SCHEDULE=30` to run more frequently when calling the related celery commands to so you can see it actually happening sooner. Feel free to update that time to whatever you like.

`CELERY_QUEUE_NAMES` is taken from the code, it's the list of celery Q's we have running, you can grab it from the yaml file or settings file.

If testing locally (ie not using pods), you may want to set the `KAFKA_LISTENER_ADDRESS_FAMILY=v4` in your environment variables.

## Running
- create an Azure Source: https://github.com/cloudigrade/cloudigrade/wiki/How-do-I-run-Sources-&-Kafka-locally-for-development#creating-a-source-and-application  
- If tracking the terminal window with the celery tasks, you should see the VM/VMSS instances being successfully loaded into `cloudigrade`
- Depending on the time value you set, you should start also seeing the following log and the VM/VMSS instances being updated (should not create additional InstanceEvents since their state has not changed yet): 
```
2022-10-18 10:44:50,532 | INFO | 54561 | strategy.py:task_message_handler:161 | requestid:None | Task api.clouds.azure.tasks.update_azure_instance_events[167c7dc5-adf0-435e-8c81-f67ebaf707b4] received
```
- In Portal (Azure), start a stopped machine or stop a running machine.
- After your specified time period, you should see the update task run and a new InstanceEvent created in the DB for that instance that matches either `power_on` or `power off` depending on what happened.

# Notes/Questions
- At the time, I have not made any changes to tests, I will update once that has been completed
- I was curious if instead of running it in a for loop for each cloud account, that I should run things asynchronously? Right now, because we're testing with 1 account, it's not a huge necessity, but as we have people using the service and adding more accounts, I think this will be important. Thoughts?
  - Async has been added

### DIFF
--- a/cloudigrade/api/tests/clouds/azure/util/test_create_initial_azure_instance_events.py
+++ b/cloudigrade/api/tests/clouds/azure/util/test_create_initial_azure_instance_events.py
@@ -36,8 +36,12 @@ class CreateInitialAzureInstanceEvents(TestCase):
         self.assertTrue(AzureInstance.objects.count() == 2)
         self.assertTrue(AzureInstance.objects.filter(resource_id=vm1["id"]).exists())
         self.assertTrue(AzureInstance.objects.filter(resource_id=vm2["id"]).exists())
-        self.assertTrue(AzureInstanceEvent.objects.count() == 1)
-        self.assertTrue(InstanceEvent.objects.count() == 1)
+        self.assertTrue(AzureInstanceEvent.objects.count() == 2)
+        self.assertTrue(InstanceEvent.objects.count() == 2)
+        # Both VMs have associated InstanceEvents created
         self.assertEqual(
-            InstanceEvent.objects.first().event_type, InstanceEvent.TYPE.power_on
+            InstanceEvent.objects.first().event_type, InstanceEvent.TYPE.power_off
+        )
+        self.assertEqual(
+            InstanceEvent.objects.last().event_type, InstanceEvent.TYPE.power_on
         )
--- a/deployment/clowdapp.yaml
+++ b/deployment/clowdapp.yaml
@@ -596,6 +596,8 @@ objects:
           value: ${CHECK_AND_CACHE_SQS_QUEUES_LENGTHS_SCHEDULE}
         - name: ENABLE_SYNTHETIC_DATA_REQUEST_HTTP_API
           value: ${ENABLE_SYNTHETIC_DATA_REQUEST_HTTP_API}
+        - name: UPDATE_AZURE_INSTANCE_EVENTS_SCHEDULE
+          value: ${UPDATE_AZURE_INSTANCE_EVENTS_SCHEDULE}
     - name: listener
       #
       # Component cloudigrade-listener deployment:
@@ -1147,6 +1149,8 @@ objects:
           value: ${CHECK_AND_CACHE_SQS_QUEUES_LENGTHS_SCHEDULE}
         - name: ENABLE_SYNTHETIC_DATA_REQUEST_HTTP_API
           value: ${ENABLE_SYNTHETIC_DATA_REQUEST_HTTP_API}
+        - name: UPDATE_AZURE_INSTANCE_EVENTS_SCHEDULE
+          value: ${UPDATE_AZURE_INSTANCE_EVENTS_SCHEDULE}
 # Secrets
 - apiVersion: v1
   data:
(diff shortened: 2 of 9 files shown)


cloudigrade/api/tests/clouds/azure/util/test_create_initial_azure_instance_events.py
```diff
@@ -36,8 +36,8 @@ def test_create_initial_azure_instance_events(self, mock_enable):
         self.assertTrue(AzureInstance.objects.count() == 2)
         self.assertTrue(AzureInstance.objects.filter(resource_id=vm1["id"]).exists())
         self.assertTrue(AzureInstance.objects.filter(resource_id=vm2["id"]).exists())
-        self.assertTrue(AzureInstanceEvent.objects.count() == 1)
-        self.assertTrue(InstanceEvent.objects.count() == 1)
+        self.assertTrue(AzureInstanceEvent.objects.count() == 2)
+        self.assertTrue(InstanceEvent.objects.count() == 2)
         self.assertEqual(
-            InstanceEvent.objects.first().event_type, InstanceEvent.TYPE.power_on
+            InstanceEvent.objects.first().event_type, InstanceEvent.TYPE.power_off
```
infinitewarp, line 42: What caused this to change? If we were asserting something about `power_on` is created before this change, should there be another test or more assertions to ensure the `power_on` is still created?

BrittRH: It's because how the test is set up originally, it's looking for one InstanceEvent with the state `power_on`, however, now that we add an InstanceEvent regardless of state, the first instance event happens to be for vm1 which is not running and in the `power_off` state. I can add another line that asserts that the last object is `power_on` (ie show that both states are covered in InstanceEvents.

BrittRH: So basically I can add another line. The first one corresponds to vm1 which is not running (thus is in a `power_off` state), and vm2 with the `power_on` state. Or I can do a count, and that there's one `power_off` InstanceEvent and one `power_on` InstanceEvent :

```python
    self.assertEqual(
        InstanceEvent.objects.first().event_type, InstanceEvent.TYPE.power_off
    )
    self.assertEqual(
        InstanceEvent.objects.last().event_type, InstanceEvent.TYPE.power_on
    )
```

infinitewarp: Yes, let's add the additional assertion to find the `power_on` event since confirming its existence was probably the original intention of that test.

BrittRH: Awesome, I've added that line to the test

infinitewarp (dismissed):

deployment/clowdapp.yaml
```diff
@@ -1819,6 +1825,10 @@ parameters:
   name: TENANT_TRANSLATOR_SCHEME
   required: true
   value: 'http'
+- name: UPDATE_AZURE_INSTANCE_EVENTS_SCHEDULE
```
abellotti, line 1828: This can be deleted as it's also defined at the bottom

abellotti (changes requested):
PR looks good, just the duplicate parameter definition.  Thanks.

abellotti (approved):

infinitewarp (approved):

abellotti (approved):

PR #1371, opened 2022-10-20
update django ~=4.1.2

infinitewarp:


infinitewarp:
After playing around more with an ephemeral deployment last week and today (`ephemeral-8i3i2v` still available and running as of this comment), I think I'm comfortable calling the Django 4.x upgrade ready. I don't anticipate any issues with this upgrade.

I deployed the current version of cloudigrade, synthesized lots of data, reran scheduled tasks to calculate runs and concurrent usage, and made API calls to look at the data. Then I built and deployed a new image with Django 4.x over that environment, checked the old data via the API, ran the tasks again, checked APIs data again, synthesized new data, and checked the APIs yet again. Everything I hit along the way appears to work normally.

The IQE PR checks have been happy the whole time, but they always operate with a new clean environment that has no preexisting data from the old version. Maybe I'm a just little paranoid, but I wanted to be extra cautious since a major 3.x to 4.x upgrade could introduce issues specifically around old migrated data. Hopefully the IQE smoke tests that run in stage (which has existing data) will catch anything I might have missed. Hopefully nothing at all. :slightly_smiling_face:

gtanzillo (approved):
LGTM 👍

Issue #1264, opened 2022-07-14
Internal api to to flush all cache (`FLUSHALL`)


infinitewarp:
## Summary

As a cloudigrade sysop, I want to be able to manually flush all ElastiCache/Redis-cached data so that I can ensure old cached values are not being used. This could be useful in case of DR (disaster recovery). For example, If we need to redeploy using a different AWS account or region, cached values for URLs in the old account may be invalid and need to be removed.

The Redis command `FLUSHALL` does this, but we should investigate and prefer a cache backend-agnostic option (at the Django cache module level) if possible so that the same API works for unit tests, etc.

## Acceptance Criteria

- [x] Verify an internal API exists that flushes all cached data

## Assumptions and Questions

- Why would we need this? What are some alternate solutions?
  - Why? We would only very rarely (maybe never!) need this, but it could be critical in case of a disaster situation such as changing our AWS account ID because some cached values may be referencing URLs or ARNs in the old AWS account.
  - Alternately, we could during that disaster contact SRE and ask them to restart all ElastiCache instances. Because the cache is in memory, restarting the instance also clears the cache. However, we don't know how long it would take to contact SRE, convince them, and get the instances restarted. By providing an internal API, we can complete the operation faster.
- We're assuming that cloudigrade is the only client for our configured ElasticCache instances. If there are other clients we don't know about, using `FLUSHALL` could destroy their data too.
  - TODO Confirm this!
- For the LocMemCache case, looks like a simple django.core.cache's cache.clear() does the trick.

infinitewarp:
Works as advertised! Thanks!

I captured a live demo to confirm. As expected, this API clears _**everything**_ from the Redis cache, and that includes not just Django-cached key-value pairs but also all Celery task queues and Celery task results. We must exercise caution around this API in stage and prod.


[![asciicast](https://asciinema.org/a/536953.svg)](https://asciinema.org/a/536953)

PR #1388, opened 2022-11-07
Adding support for /internal/clear_cache

abellotti:
- Adding support for /internal/clear_cache, this triggers a cache clear for deleting all cached keys.

For: https://github.com/cloudigrade/cloudigrade/issues/1264


abellotti:
In addition to the LocMemCache py39 tests, I've manually tested running locally as well as port forwarding to an Ephemeral deployment (both Redis).

infinitewarp (approved):

PR #1389, opened 2022-11-07
add more log messages in the Azure vm discovery process

infinitewarp:


cloudigrade/util/azure/vm.py
```diff
@@ -46,6 +54,11 @@ def get_vms_for_subscription(azure_subscription_id):
             )
             for discovered_vm in vmss_vm_list:
                 vms.append(vm_info(cm_client, image_properties, discovered_vm))
+        vmss_vms_count = len(vms) - vms_count
```
abellotti, line 57: why the `- vms_count` ? virtual_machine_scale_sets.list_all `should` return just the VMSS vms no ?

infinitewarp: `list_all` returns a custom object that acts as a generator. Once you iterate through its contents, though, it's empty. So, you'd need to either make a copy of its results in a regular list or measure some side effects like this. Since we've already iterated and built up a set, it's cheaper just to measure the difference in the vms set size.

infinitewarp: Also, vmss_list is the list of scale sets, not the vms within them.

abellotti: Ahh yes, I remember that. this is good. 👍

abellotti (approved):

PR #1392, opened 2022-11-09
add github workflow jobs to check .python-version against the actual python version provided by build-container

infinitewarp:
This adds a rudimentary test to check the installed version of Python against what's defined in `.python-version`. For now, this is only effectively testing what version is installed in the test's `build-container` image, not necessarily the `cloudigrade` image itself. I may iterate on this in a future branch+PR to try to assert the latter as well, but I think this is a good enough start.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -17,6 +17,15 @@ jobs:
       - run: tox -e py39
       - run: codecov
 
+  test-python-version:
+    name: python-version
+    runs-on: ubuntu-latest
+    container:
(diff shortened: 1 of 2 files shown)


abellotti (approved):

infinitewarp:
It you dig into the job history, [you can find a failure of this new job](https://github.com/cloudigrade/cloudigrade/actions/runs/3433799591). This is good! At the time that specific job ran, the installed Python version _did not_ match the `.python-version` definition, and the test failed as expected.

3 9 13-vs-ubi8 6-python3 9 7

BrittRH (approved):

PR #1397, opened 2022-11-14
convert TextField to JSONField where appropriate

infinitewarp:


infinitewarp:
Closing since this kind of optimization is no longer a priority.

PR #1398, opened 2022-11-14
Identity RHEL in Azure via tag

abellotti:
Identify RHEL images in Azure via the `cloudigrade-rhel-present` tag

### DIFF
--- a/cloudigrade/util/azure/vm.py
+++ b/cloudigrade/util/azure/vm.py
@@ -87,6 +87,7 @@ def vm_info(cm_client, image_properties, discovered_vm, vm_with_status=None):
         vm["running"] = is_running(vm_with_status)
     else:
         vm["running"] = is_running(discovered_vm)
+    vm["tags"] = getattr(discovered_vm, "tags", {}) or {}
     vm["is_encrypted"] = is_encrypted(discovered_vm)
     image_properties = get_image_properties(cm_client, image_properties, discovered_vm)
     vm["architecture"] = architecture(image_properties)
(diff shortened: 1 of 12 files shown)


cloudigrade/util/azure/vm.py
```diff
@@ -87,6 +87,7 @@ def vm_info(cm_client, image_properties, discovered_vm, vm_with_status=None):
         vm["running"] = is_running(vm_with_status)
     else:
         vm["running"] = is_running(discovered_vm)
+    vm["tags"] = getattr(discovered_vm, "tags", {}) or {}
```
infinitewarp, line 90: Why the ` or {}` if `getattr` already returns `{}` when `tags` is not found?

Are there cases where `tags` is present but it's not a `dict`, like does it ever appear as `None` or `False` or something else _falsy_?

abellotti: yea, I didn't have that at first. Sometimes provided to us as None. The behavior of getattr works as expected on dict but not on objects, in such cases we need the extra `or {}`.

This is expected:

```
$ python
vm={"tags": None}
print(getattr(vm, "tags", {}))
{}
```

This is not, thus the extra `or {}`

```
$ python
class Vm:
    tags = None

vm = Vm()
print(getattr(vm, "tags", {}))
None
```

infinitewarp: Ugh. Shame on Azure for returning None instead of an empty dict. Alas, your solution is fine for dealing with that. 👍

infinitewarp (approved):

PR #1400, opened 2022-11-14
Enable Azure support in Production.

abellotti:
- Enable Azure VM discovery and support in Production

Keeping as Draft for now as it:

Depends on: https://github.com/cloudigrade/cloudigrade/pull/1398

Maybe also wait on https://github.com/cloudigrade/cloudigrade/issues/1401 to be fixed before enabling.


infinitewarp:
Closing since we're not developing the Azure integration any further.

PR #1404, opened 2022-11-16
create /internal/tasks/ API to call arbitrary Celery tasks remotely and check their async results

infinitewarp:


cloudigrade/internal/views.py
```diff
@@ -504,3 +506,47 @@
...
+        return Response(
+            {"async_result_id": async_result_id, "ready": ready, "result": result}
+        )
+    except Exception as e:
+        return Response(
+            {
+                "async_result_id": async_result_id,
+                "error_class": f"{e.__class__.__module__}.{e.__class__.__name__}",
+                "error_message": str(e),
+            }
```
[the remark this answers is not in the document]
infinitewarp: Good bot!

cloudigrade/internal/views.py
```diff
@@ -504,3 +507,82 @@ def redis_raw(request):
...
+        http localhost:8000/internal/tasks/e5a666a1-dadc-4de8-8f6e-53d4a0bec3a0/
+    """
+    try:
+        async_result = AsyncResult(async_result_id)
+        ready = async_result.ready()
+        result = async_result.get() if ready else None
+        return Response(
+            {"async_result_id": async_result_id, "ready": ready, "result": result}
+        )
+    except NotRegistered as e:
```
abellotti, line 571: Can we drop the `NotRegistered` exception catch here?  Didn't find it anywhere mentioned https://docs.celeryq.dev/en/stable/reference/celery.result.html. 

I also ran a GET /internal/tasks/bogusid/ and got a:
```
{
    "async_result_id": "bogusid",
    "ready": false,
    "result": null
}
```

Also, similar ready false and result null for a task created with a bogus task_name.

infinitewarp: Hmm… I may retrace my steps to remember where I encountered `NotRegistered` because I _definitely_ encountered that when manually testing. That's why I added specific handling for it.

As for `/internal/tasks/bogusid/`, I also encountered that behavior in my manual testing. The async result ID is by convention a UUID, but nothing appears to enforce that. The result is effectively the same as asking for an async result ID that's a UUID that doesn't exist or a task that hasn't completed yet. In these cases, the effect is the same as an outside observer. The result doesn't exist until the task is complete. So, you don't know whether or not the ID you gave ever actually belonged to a task.

infinitewarp: I found my `NotRegistered` example. It can be seen by requesting a task name that our app doesn't support like this:

```
❯ http -b localhost:8000/internal/tasks/ task_name="potato"
{
    "async_result_id": "a8cfaacd-30b5-4cb7-a431-0aede58090e2"
}
```

Since `potato` isn't defined to run on a specific queue in `CELERY_TASK_ROUTES` (`potato` isn't a function _anywhere_ for that matter), the request goes to the default `celery` queue. When the Celery worker handling the default `celery` queue picks up that task, it cannot find a `potato` task function, and the worker raises (and includes in the async task result) `NotRegistered` which is then revealed when you try to get that result.

```
❯ http -b localhost:8000/internal/tasks/a8cfaacd-30b5-4cb7-a431-0aede58090e2/
{
    "async_result_id": "a8cfaacd-30b5-4cb7-a431-0aede58090e2",
    "error": "not a registered task name: potato"
}
```

abellotti: I had run a test with an invalid task_name and did not get a `NotRegistered` exception. The problem was that I did not have a worker running at the time. When a celery worker is running then the exception is properly raised.

abellotti (approved):

PR #1449, opened 2023-01-25
drop indexes, constraints, and tests around legacy models

infinitewarp:
This PR severs the schema connections to CloudAccount and User (which we want to keep) from Instance, MachineImage, ConcurrentUsage, and others, and this PR's migrations _should_ execute relatively quickly. The migrations here are designed specifically _only_ to drop constraints, not drop columns or tables, because I don't want to risk holding the service hostage during the deployment while the contents of large tables are altered or deleted. A future PR will actually drop the tables we no longer need.

View the SQL it will run by using the following commands. I have copied their outputs to gists linked below.
```
./cloudigrade/manage.py sqlmigrate api 0066
./cloudigrade/manage.py sqlmigrate api 0067
```
- [0066_alter_concurrentusage_options_and_more.py](https://github.com/cloudigrade/cloudigrade/blob/drop-indexes/cloudigrade/api/migrations/0066_alter_concurrentusage_options_and_more.py): https://gist.github.com/infinitewarp/0e465051efdfc7e742d3426a71c349a6
- [0067_remove_instance_cloud_account_and_more.py](https://github.com/cloudigrade/cloudigrade/blob/drop-indexes/cloudigrade/api/migrations/0067_remove_instance_cloud_account_and_more.py): https://gist.github.com/infinitewarp/4ba4e7d1f1ff45bbfaed27f56195891c


infinitewarp:
I have run these migrations locally multiple times against a recent full copy of the production database, and they performed as expected. Only constraints and indexes were dropped, and when run locally, all schema changes completed in less than one second.

I don't have high CPU/IO load on my system like production RDS probably has, but since I've already stripped away most application functionality around these tables, I still expect there should be no blocking or locking activity around them. 🤞

PR #1491, opened 2023-03-02
Swap startupProbe for increasing livenessProbe.initialDelaySeconds

hseljenes:
ClowdApp doesn't seem to support `startupProbe` so we'll do the crappier way of increasing `livenessProbe.initialDelaySeconds`.


jlgre (approved):
Looks good - lets wait for the integration checks to run

PR #1631, opened 2024-05-02
Update black, bonfire, gunicorn, idna, pyopenssl, and sqlparse

hseljenes:
Update some dependencies with security issues.

We're ignoring vuln 65213 as there's no "fix" since it's a bad CVE report.


jlgre (approved):
lgtm

PR #1661, opened 2024-09-23
Fix ext id related issue

ssaurav-redhat:
Ext id is not being used in authentication by cloudigrade. With this we will consider ext id in case if it is present.


jlgre (dismissed):
This looks good to me - there are still a couple linting errors but I'll leave it up to you if they're worth fixing or not.

hseljenes (approved):

PR #1664, opened 2024-10-02
Drop --cache-from in deployment/scripts/build_deploy.sh

hseljenes:
Attempt to fix deployments


deployment/scripts/build_deploy.sh
Tomboyo, line 1: Could you add `set -euo pipefail` to the top of the script? I think that even though the build fails, it still gets to the line that pushes `:latest`. The script pulled `:latest` earlier, so it's just re-pushing the same image, but it looks like this is bumping the timestamp is quay which is misleading from a debugging perspective.

Tomboyo (approved):

jlgre (approved):

